PR #2, opened 2025-12-18
Add initial Actionbase CLI

zipdoki:
Initial CLI implementation for the Hands-on Guide

depends on #5

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,6 +1,6 @@
 # Actionbase CLI
 
-The Actionbase CLI is a command-line interface for managing and maintaining Actionbase clusters.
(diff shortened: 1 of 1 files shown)


em3s:
@dosu please re-apply the labels.

em3s:
This looks good overall.

Since this is an initial setup, I suggest we merge after making just these two changes:
* Change the default port from `8081` to `9300`, to avoid common dev port conflicts and keep the Hands-on Guide setup consistent.
* Replace `--server` with a clearer `--proxy`, since the CLI proxies traffic to the Actionbase server.

We can continue refining and evolving this as the CLI grows.

Thanks for the work.

PR #17, opened 2026-01-07
scope website deploy workflow to website and workflow changes

em3s:
Restrict the website deployment workflow to run only when relevant files change.

- Trigger deployment only on changes under `website/**`
- Include `CNAME` updates
- Run when the deployment workflow itself is modified
- Keep deployment scoped to pushes on `main`

This reduces unnecessary CI runs and makes the deployment behavior clearer.


rhee519 (approved):
LGTM 👍🏻

PR #21, opened 2026-01-09
Exempt Dependabot and GitHub Actions from CLA checks

em3s:
This reverts commit f47d805c51e90b2f09f5835b0b345e938135a066.

This was not the correct solution for the CLA issue.
The issue was resolved via the GitHub App configuration, as confirmed in #18.


zipdoki (approved):
LGTM. Thanks for the quick revert.

PR #23, opened 2026-01-09
Add "Documentation" link to site social icons

em3s:
This adds a "Documentation" link to the site social icons to provide a clear path back to the docs for users entering through the blog.

The `Docs | Blog` navigation option, which would be the ideal solution, was discussed in HiDeoo/starlight-blog#190 but confirmed as not applicable for now. The guidance and considerations shared there informed this approach.

This is the most reasonable option given the current constraints.

The screenshot below shows the added Documentation link.

image


zipdoki (approved):
I was actually thinking that this feature was needed as well. Thank you for working on it.

PR #44, opened 2026-01-14
Add design context document for Actionbase

em3s:
Introduce a new "Context" document outlining the background, constraints, and adoption shaping Actionbase's design. Aimed as an entry point for understanding the system's purpose and decisions.

It will be referenced by #32.


em3s:
Closing this PR. The discussion #32 already covers enough context on its own.

PR #215, opened 2026-03-13
feat: handle NilLabel gracefully in v3 mutation API

### LINKED ISSUE #213: NilLabel WAL entries cause 400 error on AsyncProcessor replay via v3 API
## Description

When `system-mutation-mode=ASYNC` is enabled, mutations to NilLabel tables (e.g., `sys.info`) are written to WAL and queued like regular mutations. AsyncProcessor then attempts to replay these WAL entries via the v3 sync endpoint (`/edges/sync?force=true`), which rejects NilLabel because `V2BackedEngine.getTableBinding()` requires `HBaseIndexedLabel`.

**Error observed (3 occurrences during mode transition):**
```
[AbAsyncDevStreaming] InvalidHttpStatusCodeException : Failed to mutate edge:
http://ab-dev.dev.onkakao.net/graph/v3/databases/sys/tables/info/edges/sync?force=true
{"mutations":[{"type":"INSERT","edge":{"version":1773306252508,"source":"origin","target":"warmup_0","properties":{"props_active":true,"message":"warmup"}}}]}
{"status":400,"error":"Bad Request","message":"This Label (sys.info, class ...NilLabel) is not indexed or not supported for edge mutation"}
```

**Root cause:** v2 and v3 mutation APIs behave differently for NilLabel.

| Path | NilLabel support | Result |
|------|-----------------|--------|
| v2 (`/graph/v2/service/sys/label/info/edge`) | O | Returns IDLE |
| v3 (`/graph/v3/databases/sys/tables/info/edges/sync?force=true`) | X | Throws 400 |

Since AsyncProcessor relays WAL entries through the v3 API, this mismatch causes unnecessary errors during `system-mutation-mode` transitions.

## Expected Behavior

NilLabel is a no-op table that never persists data. Either:

1. **WAL should not be written for NilLabel tables** — WAL exists for data durability, but NilLabel has no data to recover. Skipping WAL at the source eliminates the problem without adding business logic to AsyncProcessor.
2. **v3 API should handle NilLabel gracefully** — align v3 behavior with v2 by returning IDLE instead of throwing 400.

eazyhozy:
## Summary

Return IDLE status for NilLabel mutations in v3 API instead of throwing
UnsupportedOperationException, aligning with v2 behavior.

Closes #213

## Changes

- Add `NilTableBinding` implementing `TableBinding` with no-op IDLE behavior
- Modify `V2BackedEngine.getTableBinding()` to return `NilTableBinding` for NilLabel
- Add `NilLabel mutation should return IDLE` test in `MutationServiceSpec`

## How to Test

```bash
./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.MutationServiceSpec"
```

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/NilTableBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/NilTableBinding.kt
@@ -27,7 +27,7 @@ class NilTableBinding(
         key: MutationKey,
         before: State,
         after: State,
-    ): Mono<MutationRecordsSummary> = Mono.just(MutationRecordsSummary(IDLE, 0, before, after))
+    ): Mono<MutationRecordsSummary> = Mono.just(MutationRecordsSummary(IDLE, 0, State.initial, State.initial))
 
     override fun handleMutationError(error: Throwable) {}
 
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/NilTableBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/NilTableBinding.kt
@@ -1,0 +1,37 @@
+package com.kakao.actionbase.v2.engine.v3
+
+import com.kakao.actionbase.core.edge.MutationKey
+import com.kakao.actionbase.core.metadata.common.ModelSchema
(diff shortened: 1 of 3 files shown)


em3s (approved):
It looks the same as v2 implementation. Nice Work!

PR #218, opened 2026-03-18
Migrate ActionbaseQuery to v3 engine layer

zipdoki:
## Summary
As discussed in https://github.com/kakao/actionbase/pull/217#pullrequestreview-3960147826, move the `/graph/v3/query` endpoint from the v2 `QueryController` to the v3 layer.
Integrate the ActionbaseQuery execution path into V3QueryService, following the existing v3 pattern (V3QueryService → Graph).

## Changes
- Move ActionbaseQuery, ActionbaseQueryExecutor, LabelProvider, Compat from v2.engine.query to engine.query                                                                
- Rename fields to v3 conventions: service→database, label→table, src→source, tgt→target, dir→direction, cache→memoize, cacheName→cache
- Consolidate QueryController to use V3QueryService instead of Graph directly                                                                                              

## Checklist
- [x] No behavioral changes — same endpoint URL, request/response format
- [x] Build and all tests pass (`./gradlew build`)


em3s (dismissed):
It should also rename v2 terms to v3 conventions – e.g. src -> source, tgt -> target in ActionbaeQuery Model.

PR #235, opened 2026-03-31
Fix missing type casting in AGG group qualifier encoding

### LINKED ISSUE #227: Missing type casting in AGG group qualifier encoding
### Bug

AGG queries return `count=0` when a group field's property type is non-String (e.g., `Long`, `Int`).

### Cause

**Write path (mutation):** Group field values are serialized using their native runtime type. For example, `1L` (Long) is encoded via `encodeInt64()` (header `0x2C`).

**Read path (AGG query):** `WherePredicate.parse(ranges)` parses the query string without schema context, so `"someLongField:eq:1"` yields a **String** `"1"`. This value reaches `ValueUtils.serialize()` without any type casting, and is encoded via `encodeString()` (header `0x34`).

The qualifier bytes differ, so the HBase lookup returns no match.

The scan path avoids this problem by applying `StartStopItem.ensureType()` to cast predicate values to the correct schema type before encoding. The AGG path needs an equivalent type casting step.

> Note: String-typed fields happen to work because both paths produce the same encoding. Fields with a `bucket` also work because `bucket.handleQueryValue().toString()` normalizes the value on both paths. These cases mask the problem.

### Steps to reproduce

1. Define a group with a non-String field:

```kotlin
mapOf(
    "group" to "my_group",
    "type" to "COUNT",
    "fields" to listOf(
        mapOf("name" to "someLongField", "bucket" to null),
        mapOf("name" to "timestamp", "bucket" to mapOf(/* date bucket */)),
    ),
)
```

2. Insert edges with `someLongField = 1L`
3. Query: `GET .../edges/agg/my_group?...&ranges=someLongField:eq:1;time:eq:...`

**Expected:** Returns the aggregated count
**Actual:** Returns `count=0`

#### Option A: Cast predicate values before encoding (minimal change)

Follow the same pattern as scan's `StartStopItem.ensureType()`

```kotlin
val eqValues = firstPairs.map { (field, predicate) ->
    require(predicate is WherePredicate.Eq)
    val castedValue = if (field.bucket == null) {
        schema.getField(field.name).type.cast(predicate.value)
    } else {
        predicate.value
    }
    field.bucketOrGet(castedValue, ceil = false)
}
```

- Pros: Small change, consistent with the existing scan pattern
- Cons: Only fixes `V3QueryService.agg()`. Future callers of `bucketOrGet()` remain unprotected.

#### Option B: Add type information to `Group.Field` (structural fix)

Make `Group.Field` type-aware so that `bucketOrGet()` itself handles casting:

```kotlin
data class Field(
    val name: String,
    val bucket: Bucket? = null,
    val type: FieldType? = null,
) {
    fun bucketOrGet(value: Any, ceil: Boolean): Any =
        bucket?.handleQueryValue(value, ceil)?.toString()
            ?: type?.cast(value)
            ?: value
}
```

- Pros: Fixes the problem at its source. All callers of `bucketOrGet()` are safe by default.
- Cons: Wider change — requires propagating type information when constructing `Group.Field` from the schema.

### References

- `V3QueryService.agg()` — parses ranges without schema
- `Group.Field.bucketOrGet()` — no type casting when `bucket` is null
- `ValueUtils.serialize()` — encodes based on runtime type
- `IndexedLabelMixin.StartStopItem.ensureType()` — scan's type casting for reference

zipdoki:
## Summary

Cast AGG predicate values using schema field types resolved at entity construction.

Closes kakao/actionbase#227

## What changed

- `Group.Field` gains a `type: PrimitiveType?` field. `bucketOrGet()` uses it to cast when no bucket is present.
- `LabelEntity` factory methods resolve `DataType` → `PrimitiveType` from the schema. Resolution always runs to avoid stale values.
- `DECIMAL` fields throw an explicit error (not yet supported in `PrimitiveType`).
- Added unit tests for `bucketOrGet()` branching and an E2E test covering bucket-only, String+bucket, and Long+bucket group configurations.

## How to test

```bash
# Unit tests — bucketOrGet() branching logic
./gradlew :core:test --tests "com.kakao.actionbase.core.metadata.common.GroupFieldTest"

# E2E tests — full AGG query flow (mutation → query) for 3 group types
./gradlew :server:test --tests "com.kakao.actionbase.server.api.graph.v3.EdgeAggQueryE2ETest"

# Full build
./gradlew build
```


zipdoki:
@em3s 
Closing in favor of a new PR implementing the revised approach from #227 — direct branching at the call site instead of carrying type on Field.

Issue #231, opened 2026-03-27
Decouple V2 dependencies from query path


em3s:
## Background

The query path (`V3QueryService`, `ActionbaseQueryExecutor`, `LabelProvider`, `CompatKt`) directly depends on 36 V2 types. The mutation path was already decoupled in #201 using the `MutationEngine` interface + `V2BackedEngine` pattern. Apply the same pattern to queries.

## Task

- [ ] Decouple like #201

## Done When

- `boundary-check`: 0 V2 edges for `ActionbaseQueryExecutor`, `LabelProvider` (deleted), `CompatKt` (deleted)
- All V2 types confined to `V2Backed*` classes
- All existing tests pass

## Notes

- V2 types in `ActionbaseQuery` itself (`DataType`, `StatKey`, `WherePredicate`) are out of scope — separate issue
- Related: 
    - #200 
    - #201

em3s:
Closing as stale.

PR #237, opened 2026-04-01
Rename V3QueryService to QueryService

em3s:
## Summary

- Rename `V3QueryService` → `QueryService`, dropping the `V3` prefix as the canonical query service
- Move package from `com.kakao.actionbase.v2.engine.v3` → `com.kakao.actionbase.engine.service` (alongside `MutationService`)
- Update all controller and test references
- Clean up `ScanFilter.Companion.defaultLimit` → `ScanFilter.defaultLimit`

Partial fix for #231

## Context

Prerequisite for #233 (QueryEngine interface extraction). Separating the rename so the subsequent refactoring diff stays clean.

## How to Test

```
./gradlew build
```


em3s:
Merging

PR #239, opened 2026-04-03
Skip write warmup in read-only mode

eazyhozy:
## Summary

Fix read-only mode warmup: skip write operations at both HTTP and engine levels instead of blocking them with the filter.

## Problem

When running with `actionbase.read-only=true` and warmup enabled, two write paths caused issues:

1. **HTTP warmup** (`ServiceLabelEdgeControllerWarmUp`): Sends POST/PUT/DELETE to `/graph/v2/service/sys/label/info/edge` on localhost — blocked by `ReadOnlyRequestFilter`
2. **Engine warmup** (`Graph.warmUp`): Performs fake edge writes to warm HBase connections — unnecessary write load on production HBase from read-only servers

## Changes

- `ServiceLabelEdgeControllerWarmUp`: Use GET-only warmup methods when `readOnly=true`
- `GraphConfig`: Add `readOnly: Boolean` field
- `GraphConfiguration`: Pass `serverProperties.readOnly` to `GraphConfig`
- `Graph.kt`: Extend existing `label.entity.readOnly` check with `|| readOnly` to skip engine-level write warmup
- `StartUpTest`: Add E2E test for `read-only=true` + `warmup.enabled=true` combination

## How to Test

```bash
./gradlew :server:test --tests "com.kakao.actionbase.server.StartUp*"
./gradlew :server:test --tests "com.kakao.actionbase.server.filter.ReadOnlyRequestFilterTest"
```

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
@@ -117,6 +117,7 @@ class Graph(
     LabelProvider,
     AutoCloseable {
     internal val mutationRequestTimeout = config.mutationRequestTimeout
+    internal val readOnly = config.readOnly
 
     private var metadataInitialized = false
 
@@ -686,7 +687,7 @@ class Graph(
                 if (it) {
                     Mono
                         .defer {
-                            if (label.entity.readOnly) {
+                            if (label.entity.readOnly || readOnly) {
                                 Mono.empty()
                             } else {
                                 writeWarmUp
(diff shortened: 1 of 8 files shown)


eazyhozy:
@em3s Gentle reminder — updated per your feedback (filter activates after warmup instead of path allowlist).

em3s:
How about just adding some condition to avoid writing? @eazyhozy . 

https://github.com/kakao/actionbase/blob/d0299f8eeb443c8cb674bcdaa9fe8c15cbf7ca27/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt#L689

em3s:
@eazyhozy 

Pushed a commit based on my earlier suggestion — extending the `if (label.entity.readOnly)` condition in `Graph.kt`. Please review and update the PR description accordingly.

eazyhozy:
@em3s Thanks for pushing the Graph.kt change — this covers the engine-level warmup I was unsure about scoping in. LGTM, updated the PR description to reflect the final approach.

Issue #247, opened 2026-04-13
Build a V3-native query engine


em3s:
## Background

`v2.engine.Graph` is a god class – early tech debt.
The V3 API path currently runs on `V2Backed*` adapters over V2 internals – a transitional shape.
This issue replaces the **query path** with a V3-native engine. Mutation is out of scope (do it later).

## Task

- [ ] Engine skeleton – Empty `Engine` Spring bean with a catalog reload loop
- [ ] Catalog reload – `Engine` periodically loads databases / tables / aliases from the metastore via a `CatalogLoader`
- [ ] Query – `Engine` produces `TableBinding` instances; `QueryService` uses it

## Done When

- V3 query controllers serve traffic through `Engine` instead of `V2BackedTableBinding`.
- `Graph` is unchanged.
- Each phase lands as one or more independently shippable PRs.

## Notes

- "Catalog" = the registry of Database / Table / Alias definitions (Spark / Trino / Iceberg / Glue use the same term). The metastore stays as the `Actionbase Table` backing store; the catalog is what lives in memory and is loaded *from* the metastore.
- Reuse `core.edge.mapper.*`, `core.edge.payload.*`, `core.metadata.common.*` — these are already V3.

em3s:
I'll handle the first task to establish the initial direction. After that, could you please take over the rest of the implementation? @zipdoki

zipdoki:
@em3s 
Sounds good! Feel free to ping here once the work is complete — happy to take a look.

em3s:
Closing as stale. We'll revisit if it comes up again.

PR #248, opened 2026-04-13
Engine skeleton with catalog reload loop

em3s:
## Summary

Phase 1 of the V3-native engine. A new `Engine` class lives in `engine.runtime` with zero `v2.*` and zero Spring imports. `PeriodicCatalogLoader` mirrors `Graph.startMetastoreReload` but the reload body is empty until phase 2.

`Graph` is unchanged. The new bean runs alongside it under `actionbase.catalog.reload-interval`.

Refs #247

## Changes

- `Engine` — composition root, `AutoCloseable`, `Engine.create(...)` factory.
- `CatalogLoader` — small interface (`bind` + idempotent `close`). Catalog = the registry of Database / Table / Alias definitions, mirroring industry terms (Spark / Trino / Iceberg / Glue).
- `PeriodicCatalogLoader` — single Reactor pipeline. `Flux.interval` for periodic, `Mono.delay(...).flux()` for one-shot. Both share the same operators, error handler, and disposable.
- `EngineConfiguration` — single `@Bean` calling `Engine.create(...)`. Only `Engine` is imported.
- `ServerProperties.CatalogProperties` — nested under the existing `actionbase` prefix.
- `application.yaml` — `actionbase.catalog.reload-interval: 1m` alongside the existing `kc.graph.metastoreReloadInterval`.
- Tests: construction binding, lifecycle delegation, periodic schedule, bind-triggers-exactly-one-reload, disabled-but-still-initial, initial-delay defers first reload.

## How to Test

```
./gradlew :engine:test --tests "com.kakao.actionbase.engine.runtime.*"
./gradlew :server:compileKotlin
```

Boot the server and look for: `Starting Flux.interval for reloading catalog every 60000 ms after 0 ms delay.`

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/Engine.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/Engine.kt
@@ -1,4 +1,7 @@
-package com.kakao.actionbase.engine.runtime
+package com.kakao.actionbase.engine
+
+import com.kakao.actionbase.engine.catalog.CatalogLoader
+import com.kakao.actionbase.engine.catalog.PeriodicCatalogLoader
 
 import java.time.Duration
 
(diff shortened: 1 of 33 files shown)


em3s:
stale

PR #245, opened 2026-04-10
Clean up v2/v3 I/O mixing on the multi-hop query path

zipdoki:
## Summary

Route the multi-hop query path (`ActionbaseQuery`) through the same v3 `TableBinding` boundary as the single-hop REST path, so nothing under `engine/query/` reaches into v2 `Label` directly anymore.

Stacked on #250 — that needs to land first. Follow-up to https://github.com/kakao/actionbase/pull/243#issuecomment-4205156057.

```
                              QueryService
                                   │
                   ┌───────────────┴────────────────┐
                   ▼                                ▼
          REST single-edge API                ActionbaseQuery
      (count / scan / get / seek / agg)          (multi-hop)
                   │                                │
                   │                                ▼
                   │                      ActionbaseQueryExecutor
                   │                                │
                   └────────────────┬───────────────┘
                                    ▼
                               TableBinding
                   (v3 boundary — DataFrame in/out,
                    base columns: version / source / target / direction)
                                    │
                                    ▼
                          V2BackedTableBinding
                   (v2 Label → v3 DataFrame, with
                    ts/src/tgt/dir → version/source/target/direction)
                                    │
                                    ▼
                             Graph.getLabel
                      (v2 direct-label access,
                       alias redirection included)
                                    │
                                    ▼
                          v2 HBaseIndexedLabel
```

- Above the boundary: v3 `DataFrame` (flat `Row` + `StructType`, base columns renamed).
- Below the boundary (only visible to the `V2Backed*` bridges): v2 `DataFrame` / `Row`, v2 `ScanFilter`, `EmptyEdgeIdEncoder`, `Direction`, etc.
- `TableBinding.agg` still returns `DataFrameEdgeAggPayload`; migrating its nested shape is out of scope.

## Test plan

- `./gradlew :engine:test` — engine unit tests including the rewritten `ActionbaseQueryExecutor*Spec` suites and `MultiHopQuerySpec`.
- `./gradlew :server:test --tests "*ActionbaseQueryE2ETest*"` — multi-hop scan+cache + scan-only E2E; asserts v3 column names (`version` / `source` / `target` / `direction`) in the JSON response.
- Manual multi-hop `ActionbaseQuery`: confirmed response base columns never leak v2 names (`ts` / `src` / `tgt` / `dir`).
- Grep audit under `engine/query/`: no production import of v2 `Label` / `LabelProvider` / `Compat` / `EmptyEdgeIdEncoder` remains outside `v2/engine/v3/V2BackedTableBinding`.

PR #250, opened 2026-04-16
Introduce v3 `DataFrame` and decouple payload construction from `TableBinding`

zipdoki:
## Summary

Decouple payload construction from `TableBinding` by introducing a v3 `DataFrame` / `Row` model under `engine/sql`. Query methods (`gets`, `count`, `scan`, `seek`) now return `Mono<DataFrame>` instead of payload types directly, and `QueryService` owns the conversion to `DataFrameEdgePayload` / `DataFrameEdgeCountPayload` — consolidating field-name constants and direction-dependent logic in one place. Follow-up to the discussion in #245.

## Test plan

- `./gradlew :engine:build` — engine compile catches signature/type breakage across `TableBinding`, `DataFrame` / `Row`, `Label.cache(offset)`, and `QueryService` call sites.
- `./gradlew :engine:test --tests "*QueryServiceSpec*"` — `count` / `gets` / `scan` build payloads via `DataFrame.toEdgePayload()`, covering `offset`, `features=total`, `ranges`, and the IN-direction flip in `scan`.
- `./gradlew :engine:test --tests "*EdgeCacheQuerySpec*"` — `QueryService.seek` pagination; `HBaseIndexedLabel.cache()` `limit + 1` / `nextOffset` across `OUT` / `IN`; simplified `V2BackedTableBinding.seek()` preserves existing behavior.
- `./gradlew :engine:test --tests "*V2BackedTableBindingTest*"` — V2 array row → V3 map row conversion via `V2BackedTableBinding.toV3()`, replacing the removed `DataFrame.rowToMap()` helper alongside the new `buildEdgeStructType()`.
- `./gradlew :engine:test --tests "*LabelSpec*" --tests "*EventualConsistencySpec*"` — `Label.cache()` / `label.scan()` unchanged for existing callers after the default-`null` `offset` parameter.
- Grep audit under `engine/` and `server/` for `Mono<DataFrameEdgePayload>` / `Mono<DataFrameEdgeCountPayload>` / `TableBinding.gets(` — confirms no production caller still expects the old payload-returning signatures.

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/service/QueryService.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/service/QueryService.kt
@@ -11,8 +11,6 @@ import com.kakao.actionbase.engine.query.ActionbaseQuery
 import com.kakao.actionbase.engine.sql.DataFrame
 import com.kakao.actionbase.v2.core.metadata.Direction
 import com.kakao.actionbase.v2.engine.sql.ScanFilter
-import com.kakao.actionbase.v2.engine.v3.V2BackedTableBinding.Companion.COUNT_FIELD
-import com.kakao.actionbase.v2.engine.v3.V2BackedTableBinding.Companion.toV3
 
 import reactor.core.publisher.Mono
 
@@ -61,7 +59,7 @@ class QueryService(
                         EdgeCountPayload(
                             start = row[EdgeField.SOURCE] as Any,
                             direction = direction.toV3(),
-                            count = row[COUNT_FIELD] as Long,
+                            count = row[DataFrame.COUNT_FIELD] as Long,
                             context = emptyMap(),
                         )
                     }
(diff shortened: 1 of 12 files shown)


em3s:
The original `DataFrame` was V2 Kotlin-based. An experimental V3 version used Java + Immutables but over-generalized the Row type. This is a clean Kotlin rewrite — Row type is Map only, intentionally kept simple.

engine/src/main/kotlin/com/kakao/actionbase/engine/service/QueryService.kt
```diff
@@ -1,14 +1,18 @@
...
 import com.kakao.actionbase.core.edge.payload.EdgeCountPayload
+import com.kakao.actionbase.core.edge.payload.EdgePayload
 import com.kakao.actionbase.engine.QueryEngine
 import com.kakao.actionbase.engine.query.ActionbaseQuery
+import com.kakao.actionbase.engine.sql.DataFrame
 import com.kakao.actionbase.v2.core.metadata.Direction
-import com.kakao.actionbase.v2.engine.sql.DataFrame
 import com.kakao.actionbase.v2.engine.sql.ScanFilter
+import com.kakao.actionbase.v2.engine.v3.V2BackedTableBinding.Companion.COUNT_FIELD
+import com.kakao.actionbase.v2.engine.v3.V2BackedTableBinding.Companion.toV3
```
em3s: This layer should not reference the implementation layer.

Please remove all V2 references if possible — except for `com.kakao.actionbase.v2.engine.sql.ScanFilter`.

zipdoki: @em3s 
Done, removed all of them. Thanks for the catch!

em3s:
Nice Work! @zipdoki 

Optimistic Merging.

PR #252, opened 2026-04-21
Backport read-only mode to 0.2.x

eazyhozy:
## Summary

Backport read-only mode from \`main\` to the \`0.2.x\` release branch, so it can be included in a v0.2.1 patch release.

## Changes

- Cherry-pick d936f56 — Add read-only mode to reject write operations (#230)
- Cherry-pick f0a84e8 — Skip write warmup in read-only mode (#239)
- Add \`POST /graph/v2/query\` to \`READ_ENDPOINTS\` in \`ReadOnlyRequestFilterTest\` (0.2.x-specific: this endpoint was removed on \`main\` via #228 but still exists on 0.2.x; the filter's \`readSuffixes\` already includes \`/query\` so no filter logic change is needed)

Note: PR #236 (*Fix ReadOnlyRequestFilterTest for renamed edges/cache → edges/seek endpoint*) is intentionally **not** included — the \`edges/cache → edges/seek\` rename has not been applied to 0.2.x, so applying #236 would break the test suite here.

## How to Test

```bash
./gradlew :server:test --tests "*ReadOnly*" --tests "*StartUp*"
./gradlew build
```

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
@@ -116,6 +116,7 @@ class Graph(
     LabelProvider,
     AutoCloseable {
     internal val mutationRequestTimeout = config.mutationRequestTimeout
+    internal val readOnly = config.readOnly
 
     private var metadataInitialized = false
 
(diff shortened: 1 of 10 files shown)


eazyhozy:
@em3s 

CI was not triggered because the base branch is `0.2.x` instead of `main`. Verified locally: all 269 server tests pass with no failures.

```
./gradlew :server:test --rerun-tasks
→ 269 tests completed, 0 failures, 0 errors — BUILD SUCCESSFUL
```

PR #254, opened 2026-04-21
Capture `Put`/`Delete`/`Increment` into `StorageOpCollector`

### LINKED ISSUE #253: Expose StorageBackend operation details in mutation response
## Background

Shadow tests compare mutations between production and test environments, but only on response fields (`source`, `target`, `status`, `count`). We can't see which put/delete/increment operations were actually issued to storage — comparison stays shallow.

## Task

- [x] #258 — `context` slot on V2/V3 mutation response items
- [x] #259 — opt-in header + e2e wire verification
- [x] #261 — storage-op collector scaffold (wiring + API)
- [ ] #254 — storage-op capture implementation + tests

## Done When

- With the opt-in header, mutation responses list the put/delete/increment ops actually issued
- Without the header, wire format is byte-identical to today
- Shadow tests can compare at the operation level

em3s:
## Summary

Closes #253. `StorageOpCollector.collect(...)` now decodes HBase `Put` / `Delete` / `Increment` into base64-encoded `StorageOp`. No call-site changes.

Set the `AB-Include-Mutation-Context: true` header on a mutation request to receive the captured ops under `context["storageOps"]` in the response. When the captured set exceeds the `DEFAULT_MAX_OPS` cap, `context["storageOpsTruncated"] = true` is added so clients can tell a truncated snapshot from a complete one. Without the header the wire format is unchanged.

## Changes

- `StorageOp` — adds `Put` / `Delete` / `Increment` / `Unknown` data classes; `Cell` carries `type` (`Put`, `DeleteColumn`, `DeleteFamily`, `DeleteFamilyVersion`, `Delete`) so delete subtypes are observable
- `StorageOpCollector.collect` — `when` on `Put` / `Delete` / `Increment`; base64-encode row/family/qualifier/value via `ByteBuffer.wrap` to avoid an extra cell clone; unknown request types become `StorageOp.Unknown` with a one-shot per-type `warn` log
- `StorageOpCollector.toContextMap()` — single source of the `storageOps` / `storageOpsTruncated` shape, consumed by V3 `MutationService` and V2 `Graph`
- `CdcContext.storageOpsTruncated` — carries the flag across the V2 pipeline; `@JsonIgnore` keeps it out of CDC

## How to Test

`./gradlew build`

- `StorageOpCollectorTest` — round-trip per op type, multi-CF Put, multi-byte value, truncation, `toContextMap` flag, `Append` → `Unknown`, non-`Mutation` fallthrough
- `CdcContextSerializationTest` — `storageOps` and `storageOpsTruncated` excluded from CDC
- `V2EdgeStorageOpsSpec` — V2 `/graph/v2/edge` with header
- `EdgeMutationStorageOpsSpec` — V3 edge + multi-edge, `context.storageOps[0]` assertions

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/storage/StorageOp.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/storage/StorageOp.kt
@@ -1,3 +1,50 @@
 package com.kakao.actionbase.core.storage
 
-sealed class StorageOp
+sealed class StorageOp {
+    abstract val kind: String
+    abstract val table: String
+
+    data class Put(
+        override val table: String,
+        val row: String,
+        val cells: List<Cell>,
+    ) : StorageOp() {
+        override val kind: String = "Put"
+    }
+
+    data class Delete(
+        override val table: String,
+        val row: String,
+        val cells: List<Cell>,
+    ) : StorageOp() {
+        override val kind: String = "Delete"
+    }
+
+    data class Increment(
+        override val table: String,
+        val row: String,
+        val deltas: List<Delta>,
+    ) : StorageOp() {
+        override val kind: String = "Increment"
+    }
+
+    data class Unknown(
+        override val table: String,
+        val type: String,
+    ) : StorageOp() {
+        override val kind: String = "Unknown"
+    }
+
+    data class Cell(
+        val family: String,
+        val qualifier: String,
+        val value: String?,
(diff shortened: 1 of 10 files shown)


em3s:
This PR went through several iterations — split across 

- #258 
- #259 
- #261 

— before landing here.

em3s:
Optimistic Merge.

PR #258, opened 2026-04-22
Add `context` map to mutation response payloads

em3s:
## Summary

Query payloads (`EdgePayload`, etc.) already carry a `context: Map<String, Any?>?` slot; add the same slot to mutation payloads for symmetry. Applies to both V2 and V3 endpoints.

Nullable with `@JsonInclude(NON_NULL)` — **non-breaking**. Existing callers are unaffected.

## Changes

- `core.MutationResult.context: Map<String, Any?>? = null`
- V3 `EdgeMutationResponse.Item.context` and `MultiEdgeMutationResponse.Item.context`
- V2 `engine.edge.MutationResultItem.context`
- `from(...)` factories thread it through unchanged

## Tests

- `MutationResponseContextTest` (core) — V3 items ser/der: null omitted, non-null round-trip
- `MutationResultItemContextTest` (engine) — same for V2 `MutationResultItem`


em3s:
Merging

PR #259, opened 2026-04-22
Verify mutation response `context` is non-breaking via opt-in header and e2e

em3s:
## Summary

Follows #258. Adds an opt-in header `AB-Include-Mutation-Context` (case-insensitive) that attaches an empty `context: {}` map to each mutation response item. Without it, `context` stays `null` and is omitted — wire-identical for existing callers. Covers V2 and V3 endpoints, verified end-to-end.

## How it works

V3 carries the flag on `RequestContext.includeContext`, which the existing `ServerRequestContextArgumentResolver` already assembles from request headers — V3 controllers stay untouched. V2 `EdgeController` also receives `RequestContext` and forwards the flag to `Graph`. Other V2 service-label controllers keep the default (`false`) via `Graph`'s default parameter.

Enrichment (`.copy(context = emptyMap())`) happens *after* `.collectList()`, so WAL, storage, and CDC payloads are unaffected.

## Changes

- `HttpHeaderConstants.INCLUDE_MUTATION_CONTEXT` — new header constant
- `engine.context.RequestContext.includeContext` — new `Boolean` field, `@field:JsonIgnore` to keep it out of `GET /graph` response
- `ServerRequestContextArgumentResolver` — reads the header (trim + case-insensitive `"true"`)
- `engine.service.MutationService.mutate` — enriches items when `requestContext.includeContext`
- `v2.engine.Graph.mutate` + `upsert`/`update`/`delete` overloads — `includeContext: Boolean = false` parameter threading into the same enrichment
- V2 `EdgeController` — now takes `RequestContext` (no new `@RequestHeader` noise) and forwards `includeContext`

## Tests

- `MutationContextOptInE2ETest` (e2e via `WebTestClient`) — V3 edge sync/async + multi-edge sync, V2 edge; header on/off/false, case-insensitive (8 cases)

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/context/RequestContext.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/context/RequestContext.kt
@@ -1,8 +1,12 @@
 package com.kakao.actionbase.engine.context
 
+import com.fasterxml.jackson.annotation.JsonIgnore
+
 data class RequestContext(
     val requestId: String,
     val actor: String,
+    @field:JsonIgnore
+    val includeContext: Boolean = false,
 ) {
     companion object {
         val DEFAULT = RequestContext("-", "-")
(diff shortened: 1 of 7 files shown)


em3s:
Merging

PR #282, opened 2026-04-27
Fix V2DataFrame.toV3 crash on JSON/DECIMAL field types

zipdoki:
## Summary

`V2DataFrame.toV3()` mapped each V2 field type to the V3 `PrimitiveType` by name with `PrimitiveType.valueOf(field.type.name)`. This worked for the primitive numeric and string types whose names happen to match, but threw `IllegalArgumentException` whenever a V2 schema contained a `JSON` or `DECIMAL` column — neither name exists in `PrimitiveType`.

Replace the by-name lookup with an explicit `V2DataType.toV3PrimitiveType()` extension. The mapping mirrors the one already used by `V3MetadataConverter`, collapsing both `JSON` and `DECIMAL` to `PrimitiveType.OBJECT`. The extension is colocated with `toV3` in `V2BackedTableBinding.kt`.

## Test plan

- [x] `./gradlew :engine:compileKotlin :server:compileKotlin` passes
- [x] New `V2DataFrameToV3` test class in `V2BackedTableBindingTest` covers all 12 conversion paths, including the JSON/DECIMAL → OBJECT regression guard that previously crashed
- [x] `./gradlew :engine:test --tests V2BackedTableBindingTest` — all 30 tests green

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
@@ -579,6 +579,8 @@ internal fun V2DataFrame.toV3(total: Long? = null): DataFrame {
     )
 }
 
+// TODO: dedupe with V3MetadataConverter.toV3PrimitiveType once a shared
+// home for V2<->V3 type conversions exists (engine can't depend on server).
 internal fun V2DataType.toV3PrimitiveType(): PrimitiveType =
     when (this) {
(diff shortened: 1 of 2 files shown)


zipdoki:
@em3s
`V2DataFrame.toV3()` crashes on `JSON`/`DECIMAL` columns — `PrimitiveType.valueOf(name)` blows up because neither name exists in `PrimitiveType`. Opened this PR with the fix + regression tests. PTAL?

em3s:
Good 👍

Optimistic Merging.

PR #283, opened 2026-04-27
Work around V2 user properties shadowing V3 system fields in toV3

zipdoki:
## Summary

When a V2 user-defined property has a name that matches a V3 system field (`version`, `source`, `target`, `direction`), the V2→V3 DataFrame conversion silently overwrites the edge metadata with the user value, so callers see the wrong `version`/`source`/`target` on the edge.

Wrap colliding user property names with backticks (e.g. `` `version` ``) at the V2→V3 conversion boundary so both the system and user values remain accessible inside the V3 row map. The backticks are an internal disambiguation: at the `EdgePayload` boundary they are stripped, since `EdgePayload.properties` is its own namespace separate from the `version`/`source`/`target` fields. Callers see the natural property names.

This is a stop-gap. The proper fix is a nested row form in V3 DataFrame that gives user properties their own namespace; once that lands the escape can be removed. Multi-hop on labels with colliding property names is still unsupported.

## Test plan

- `./gradlew :engine:test --tests V2BackedTableBindingTest` — unit coverage for `V2DataFrame.toV3`: colliding user fields get backticked keys in the V3 row, edge fields keep their natural names, non-colliding fields are untouched.
- `./gradlew :engine:test --tests QueryServiceSpec` — e2e coverage on a label whose schema declares `version`/`source`/`target` as user properties: `get`, `count`, and `scan` return the correct edge metadata and surface user values under their natural keys (no backticks leak to `EdgePayload.properties`).

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
@@ -544,12 +544,20 @@ class V2BackedTableBinding(
     }
 }
 
+/**
+ * Converts a V2 DataFrame to a V3 DataFrame.
+ *
+ * User property fields whose names match V3 system fields (`version`, `source`, `target`,
+ * `direction`) are wrapped in backticks (e.g. `` `version` ``) so they do not shadow the
+ * edge metadata at those keys. This is a workaround; remove it once V3 DataFrame supports
+ * a nested row form that gives user properties their own namespace.
+ */
 internal fun V2DataFrame.toV3(total: Long? = null): DataFrame {
     val v3Schema =
         com.kakao.actionbase.core.metadata.common.StructType(
             schema.fields.map { field ->
                 com.kakao.actionbase.core.metadata.common.StructField(
-                    name = EdgeField.toV3(field.name),
+                    name = escapeIfCollides(field.name),
                     type = field.type.toV3PrimitiveType(),
                     comment = field.desc,
                     nullable = field.isNullable,
@@ -562,7 +570,7 @@ internal fun V2DataFrame.toV3(total: Long? = null): DataFrame {
                 data =
                     schema.fieldNames
                         .mapIndexed { i, name ->
-                            EdgeField.toV3(name) to row.array[i]
+                            escapeIfCollides(name) to row.array[i]
                         }.toMap(),
                 schema = v3Schema,
             )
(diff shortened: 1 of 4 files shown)


engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
```diff
@@ -544,12 +544,20 @@ class V2BackedTableBinding(
...
+ * edge metadata at those keys. This is a workaround; remove it once V3 DataFrame supports
+ * a nested row form that gives user properties their own namespace.
+ */
 internal fun V2DataFrame.toV3(total: Long? = null): DataFrame {
     val v3Schema =
         com.kakao.actionbase.core.metadata.common.StructType(
             schema.fields.map { field ->
                 com.kakao.actionbase.core.metadata.common.StructField(
-                    name = EdgeField.toV3(field.name),
+                    name = escapeIfCollides(field.name),
```
em3s: How about `escapeV3Keyword`

zipdoki:
@em3s 
Quick workaround for a V2→V3 collision bug (user props shadowing edge system fields like version/source/target). PTAL when you get a chance 🙏

engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
```diff
@@ -579,6 +587,11 @@ internal fun V2DataFrame.toV3(total: Long? = null): DataFrame {
     )
 }
 
+private val V3_SYSTEM_FIELD_NAMES =
+    setOf(EdgeField.VERSION, EdgeField.SOURCE, EdgeField.TARGET, EdgeField.DIRECTION)
+
+private fun escapeIfCollides(v2Name: String): String = if (v2Name in V3_SYSTEM_FIELD_NAMES) "`$v2Name`" else EdgeField.toV3(v2Name)
```
em3s: Please add the corresponding `unescapeV3Keyword` here as well — I've suggested renaming `escapeIfCollides` to `escapeV3Keyword`, so the pair would be `escapeV3Keyword` / `unescapeV3Keyword`.

em3s:
V2 handles properties as flat variables, so `ts` is mapped to its V3 name `version`. This causes a collision with `properties.version`, which also resolves to `version` — they can overwrite each other. This is a workaround for that.

zipdoki:
Thanks @em3s! Renamed to escapeV3Keyword and added the paired unescapeV3Keyword as suggested.

em3s:
Optimistic Merging

Issue #271, opened 2026-04-24
Migrate Kotlin data-driven tests to `@ObjectSource`


em3s:
Migrate Kotlin tests in `core` / `engine` / `server` to the data-driven family (`@ObjectSource` / `@TableSource`, see #268, #273). Java files and Kotest spec files are tracked separately.

## Phases

| Phase | Scope | PR | Status |
|---|---|---|---|
| 0 | Pilot — one file per shape | #291 | merged |
| 1 | Already-parameterized (`@CsvSource` / `@MethodSource`) | #292 | merged |
| 2 | `core` non-parameterized (Serialization + EdgeMutationStrategy + sort variants) | #295 | merged |
| 3 | `engine` + `server` stragglers | — | out of scope for now; revisit later |

## Decision rule

Migrate when multiple `@Test` methods share assertion shape *and* the table form makes the design contract more readable. Otherwise leave as `@Test`. Forcing single-scenario or heterogeneous tests through the family adds string dispatch / schema gymnastics without revealing intent.

Findings carried across phases:
- Resolver fix in #291 unblocks constructor-injected tests (HBase, Spring).
- ktlint requires backtick-named methods to live alongside `@Test` / `@Nested` — keep one of those imports in migrated files.
- Pipe cells stay raw `String` (only `~` → null); cells containing `:` need the whole row single-quoted.
- Use JSON-flow YAML for descriptors that travel as JSON over HTTP. `PrettyObjectWriter.DEFAULT` replaces the repeated indent=2 / line=80 constructor.

## Out of scope (separate issues / tracks)

- 39 Java files (needs `-parameters` + Java-reflection fallback in `ObjectSourceExtension`).
- 49 non-convertible files (Graph/HBase fixtures, contract suites, Spring/E2E).
- Kotest removal from engine spec files.
- AssertJ removal from server tests.
- `*Spec.kt` → `*Test.kt` rename.
- Rename `ObjectSourceExtension` now that it handles both sources.

## Acceptance

- [x] `rg "@CsvSource|@ValueSource" core engine server` (Kotlin-only paths) returns 0
- [x] `./gradlew test` green
- [x] Test count preserved per PR (documented in each PR body)

em3s:
Driving this with Claude Code.

em3s:
Closing — landed via #291, #292, and #295. Further refactoring will move to separate issues.

PR #293, opened 2026-04-28
Add root-list and pipe-delimited row shapes to @TableSource

em3s:
## Summary

Simplifies `@TableSource` to a single canonical shape: the annotation value is a YAML list of rows; cells map to test method parameters by position via Kotlin reflection. The previous map-with-`columns:`-and-`rows:` wrapper is removed — `columns:` was always redundant with the method signature, and the `rows:` wrapper added a layer that didn't earn its keep.

This is a precondition for the Phase 1 migration in #271 / #292: state-machine tests like `SingleTransitionTest` (41 × 32) read poorly without column-by-column scannability.

## Changes

Each row item is one of two shapes:

**1. Flow list** `[a, b, c]` — concise for short, simple tables.

```kotlin
@TableSource("""
    - [IDLE,    START, RUNNING]
    - [RUNNING, STOP,  IDLE]
""")
fun `transition`(from: State, event: Event, expected: State) { ... }
```

**2. Pipe-delimited string** `a | b | c` — recommended for wide matrices where column-by-column scannability matters. Each cell is parsed as a YAML scalar, so `~` is null, bare numbers are typed, and single-quoted strings keep their content. YAML strips `#` comments and blank lines naturally.

```kotlin
@TableSource("""
    # state machine transitions
    - IDLE    | START | RUNNING
    - RUNNING | STOP  | IDLE
""")
fun `transition`(from: State, event: Event, expected: State) { ... }
```

## Implementation

- `core/src/testFixtures/.../params/ObjectSourceExtension.kt` — `parseTableSource` now takes `methodParameters: List<String>` (supplied from reflection at call time). The annotation value parses as a YAML list directly; each row item is dispatched to `parseRow` which handles `List` (flow) or `String` (pipe).
- `core/src/testFixtures/.../params/TableSource.kt` — KDoc updated.
- `core/src/test/.../test/TableSourceTest.kt` — covers both shapes plus error paths.

## Phase 0 callers updated

The two callers that already used `@TableSource` are converted to the new shape:

- `engine/.../experiments/hbase/PerCellTTLTest.kt` (14 rows × 3 cols)
- `server/.../v3/metadata/MetadataStatusTest.kt` (6 rows × 3 cols)

Both retain their original test counts.

## Breaking change

The legacy map-with-`columns:`-and-`rows:` shape no longer parses. Any consumer outside this PR that uses `@TableSource` must adopt the new shape; #292 (Phase 1 migration) will rebase on this PR.

## How to Test

```bash
./gradlew :core:test --tests '*TableSourceTest'                  # 12 cases
./gradlew :server:test --tests '*MetadataStatusTest'             # 6 cases
./gradlew :engine:test --tests '*PerCellTTLTest'                 # 14 cases (HBase)
./gradlew :core:test :server:test                                # full sweep, no regression
./gradlew spotlessKotlinCheck
```

## AI Assistance

  - Tool / model: Claude Code with Claude Opus 4.7 (1M context)

### DIFF
--- a/core/src/test/kotlin/com/kakao/actionbase/test/TableSourceTest.kt
+++ b/core/src/test/kotlin/com/kakao/actionbase/test/TableSourceTest.kt
@@ -15,14 +15,12 @@ class TableSourceTest {
     @ObjectSourceParameterizedTest
     @TableSource(
         """
-        columns: [number, string]
-        rows:
-          - [1, foo]
-          - [2, bar]
-          - [3, baz]
+        - [1, foo]
+        - [2, bar]
+        - [3, baz]
         """,
     )
-    fun `columns become parameter names`(
+    fun `flow rows bind by parameter position`(
         number: Int,
         string: String,
     ) {
@@ -40,10 +38,8 @@ class TableSourceTest {
     @ObjectSourceParameterizedTest
     @TableSource(
         """
-        columns: [a, b, c]
-        rows:
-          - [1, 2, 3]
-          - [4, ~, 6]
+        - [1, 2, 3]
+        - [4, ~, 6]
         """,
     )
     fun `tilde maps to null for nullable parameters`(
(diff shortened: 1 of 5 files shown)


em3s:
Reviewed. Merging

PR #302, opened 2026-05-06
Add TypedKeyFieldValue and fix Cache.Field in `codec-java`

zipdoki:
## Summary

Two fixes to V2 codec edge encoding.

**`Cache.fields` wire shape.** Server `LabelEntity` mixes V2 `Index` with V3 `Cache`, so `caches[].fields[].*` uses the V3 key `field`, not V2's `name`. V2's `Cache` reused `Index.Field`, so real responses deserialized to null names and broke encoding. New `Cache.Field` annotated `@JsonProperty("field")`; `Index.Field` keeps `name`.

**Row type tag on bulk output.** `bulkEncodeAll` consumers couldn't tell hash from indexed/cache/counter rows without re-decoding keys. `TypedKeyFieldValue<T>` carries `EncodedEdgeType` with `key`/`field`/`value`, and `bulkEncodeAll` now returns `List<TypedKeyFieldValue<T>>`. Kept separate from `KeyFieldValue` so engine scan paths (`HBaseHashLabel`, `JdbcHashLabel`) stay untouched; overloaded `from` adapters bridge them.

Closes #

## Changes

- Add `Cache.Field` with `@JsonProperty("field")` to match the V3 server wire shape (`core.metadata.common.IndexField`); `Index.Field` keeps the legacy `name` key.
- Introduce `TypedKeyFieldValue<T>` carrying `EncodedEdgeType` alongside `key`/`field`/`value`.
- Switch `BulkEdgeEncoder.bulkEncodeAll` to return `List<TypedKeyFieldValue<T>>`; update `AbstractEdgeEncoder` signatures accordingly.
- Add overloaded `from` adapters between `KeyFieldValue` and `TypedKeyFieldValue`; existing engine scan/query paths are unchanged.
- Expand the wildcard metadata import in `BulkEdgeEncoder` and apply spotless formatting to the codec-java module.
- Tests: add `LabelDTOTests`; update `BulkEdgeEncoderTests` and `MultiEdgeBulkEdgeEncoderTests` for the new return type.

## How to Test

- `./gradlew :codec-java:test` — covers both fixes via `BulkEdgeEncoderTests`, `MultiEdgeBulkEdgeEncoderTests`, and the new `LabelDTOTests`.
- `./gradlew test` — full suite green; engine integration tests confirm the existing `KeyFieldValue` scan path is untouched.

## AI Assistance

### DIFF
--- a/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/KeyFieldValueV2.java
+++ b/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/KeyFieldValueV2.java
@@ -38,7 +38,12 @@ public class KeyFieldValueV2<T> {
     return value;
   }
 
-  public static <T> KeyFieldValueV2<T> fromV1(KeyFieldValue<T> keyFieldValue, EncodedEdgeType encodedEdgeType) {
-    return new KeyFieldValueV2<>(encodedEdgeType, keyFieldValue.getKey(), keyFieldValue.getField(), keyFieldValue.getValue());
+  public static <T> KeyFieldValueV2<T> fromV1(
+      KeyFieldValue<T> keyFieldValue, EncodedEdgeType encodedEdgeType) {
+    return new KeyFieldValueV2<>(
+        encodedEdgeType,
+        keyFieldValue.getKey(),
+        keyFieldValue.getField(),
+        keyFieldValue.getValue());
   }
 }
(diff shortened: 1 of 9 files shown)


codec-java/src/main/java/com/kakao/actionbase/v2/core/code/KeyFieldValueV2.java
```diff
@@ -0,0 +1,49 @@
+package com.kakao.actionbase.v2.core.code;
+
+import com.kakao.actionbase.v2.core.metadata.EncodedEdgeType;
+
+public class KeyFieldValueV2<T> {
```
em3s: How about `TypedKeyFieldValue` instead of the ambiguous V2 suffix?

zipdoki:
@em3s 
Renamed to `TypedKeyFieldValue`, and folded fromV1 / fromV2 into a single from for the same reason. Thanks!

em3s:
LGTM 👍
Merging

PR #304, opened 2026-05-07
Add ranges to the cache seek API

### LINKED ISSUE #303: Add `ranges` to the cache `seek` API
## Problem

The cache `seek` API only takes `(start, direction, limit, offset)`. For a compound cache like `(permission ASC, createdAt DESC)`, callers cannot constrain the result to a specific leading-field value — a client wanting `permission='others'` either pages through earlier categories or filters client-side, both of which break pagination. This is asymmetric with `scan` (index), which already accepts `ranges`.

## Proposed Solution

Add `ranges` to `seek`, mirroring `scan`'s grammar (`field:op:value;...`):

```
GET /…/edges/seek/permission_created_at_desc
    ?start=1&direction=OUT&limit=20
    &ranges=permission:eq:others;createdAt:lt:1700000000000
```

Predicates are translated to a single `ColumnRangeFilter` start/stop, following the physical constraints of HBase wide-row scans:

1. Predicates must form a contiguous leading prefix of the cache's `fields`.
2. Only the trailing predicate may be a range; preceding predicates must be `=`.
3. Per-field `Order` is honoured by `OrderedBytes` (DESC fields are bit-inverted at write time), so a single forward column scan already returns rows in the intended compound order. The encoding layer is unchanged; only the read path adds predicate→byte-range conversion.

Plumbed through `QueryService.seek` → `TableBinding.seek` → `Label.cache` → `HBaseIndexedLabel.cache`. Validation lives at the `TableBinding` boundary (matching `scan`); byte-range encoding lives at the label, mirroring `IndexedLabelMixin.createStartStop` and reusing the same `plusOne` convention for prefix bounds.

## Alternatives Considered

* **Use index instead of cache for filterable queries.** Loses the single-Get advantage on a source-pinned wide row; for multi-hop traversals this means one Scan per hop instead of one Get per hop.
* **Per-leading-value cache** (separate cache per `permission`). Solves filtering but causes schema sprawl and write amplification.
* **Client-side filtering.** Breaks `limit` semantics — the client cannot guarantee N items per call without unbounded overfetch.
* **Support `IN` predicates.** Would require disjoint qualifier byte ranges (multiple `ColumnRangeFilter`s OR'd, or fan-out Gets), and complicates the offset cursor. Index `scan` does not support `IN` at the byte-range level either; deferred to keep parity. A future `filters` (post-decode) channel is the natural place for it.

## Additional Context

* **Top-N eviction interaction (orthogonal but worth flagging).** `cache.limit` is declared in the schema but not currently enforced by the mutation path. If per-source top-N is added later, the sort key becomes the eviction priority — leading-categorical compound caches like `(permission ASC, …)` would risk starving trailing groups (e.g. `permission='others'` evicted in favour of `'me'`). Schema docs should call this out when eviction lands.
* **Tests.** Added `EdgeCacheRecordMapperTest` cases that (a) verify mixed-direction compound qualifiers sort lex into the intended `(leading ASC, trailing DESC)` order, and (b) verify the predicate prefix start/stop bracket matching qualifiers and exclude others — proving the byte-range used by the seek API.

zipdoki:
## Summary

Adds a `ranges` query parameter to the cache `seek` API. Predicates form a leading prefix of the cache's fields and translate to a single HBase `ColumnRangeFilter` over the qualifier bytes. Per-field `Order` is honored by `OrderedBytes` (DESC fields are bit-inverted at write time), so the forward column scan yields the declared compound order with no scan-direction handling required.

Closes #303

## Changes

- Plumb `ranges: String?` from `EdgeQueryController` and `ActionbaseQuery.Item.Seek` through `QueryService.seek` → `TableBinding.seek` → `Label.cache` → `HBaseIndexedLabel.cache`.
- Validate predicates at `V2BackedTableBinding.seek` (matching `scan`): leading-prefix order, trailing-only range, and `IN` rejection.
- Add `HBaseIndexedLabel.encodeCacheRange` to resolve predicates plus offset into one `(from, to)` byte range; offset, when present, supersedes the predicate-derived `from`.
- Add `QualifierStartStop`/`QualifierStartStopItem` as the cache-side analogue of `IndexedLabelMixin.StartStop`/`StartStopItem` (collapsible in a follow-up).
- Drop the `order` parameter and the `from`/`to` swap from `hbaseGetWideRow`; bytes are byte-ordered upstream, so the swap was a no-op.
- Add `ByteArrayBuffer.plusOne` (Kotlin port of `EdgeBuffer.plusOne`) for exclusive boundary bytes on the cache encoder path.

## How to Test

- `./gradlew :core:test --tests "com.kakao.actionbase.core.edge.mapper.EdgeCacheRecordMapperTest"`
- `./gradlew :server:test --tests "com.kakao.actionbase.server.api.graph.v3.EdgeCacheQueryE2ETest"`
- `./gradlew spotlessCheck`

## AI Assistance

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/codec/ByteArrayBuffer.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/codec/ByteArrayBuffer.kt
@@ -20,4 +20,22 @@ fun <T> ByteArrayBuffer.getValueOrNull(): T? = ValueUtils.deserialize(buffer = t
 
 fun ByteArrayBuffer.hasRemaining(): Boolean = remaining > 0
 
+fun ByteArrayBuffer.plusOne(): ByteArrayBuffer {
+    val position = this.position
+    val bytes = this.bytes
+    var carry = true
+    var i = position - 1
+    while (i >= 0 && carry) {
+        if ((bytes[i].toInt() and 0xFF) == 0xFF) {
+            bytes[i] = 0
+        } else {
+            bytes[i]++
+            carry = false
+        }
+        i--
+    }
+    require(!carry) { "Overflow while incrementing byte buffer" }
+    return this
+}
+
 fun ByteArray.buffer(): ByteArrayBuffer = ByteArrayBuffer(this)
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/mapper/EdgeCacheRecordMapper.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/mapper/EdgeCacheRecordMapper.kt
@@ -10,8 +10,10 @@ import com.kakao.actionbase.core.codec.buffer
 import com.kakao.actionbase.core.codec.getValue
 import com.kakao.actionbase.core.codec.getValueOrNull
 import com.kakao.actionbase.core.codec.hasRemaining
+import com.kakao.actionbase.core.codec.plusOne
 import com.kakao.actionbase.core.codec.putValue
 import com.kakao.actionbase.core.edge.record.EdgeCacheRecord
+import com.kakao.actionbase.core.java.codec.common.hbase.Order
 import com.kakao.actionbase.core.metadata.common.Direction
 import com.kakao.actionbase.core.storage.HBaseRecord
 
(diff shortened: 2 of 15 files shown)


core/src/main/kotlin/com/kakao/actionbase/core/codec/ByteArrayBuffer.kt
```diff
@@ -20,4 +20,22 @@ fun <T> ByteArrayBuffer.getValueOrNull(): T? = ValueUtils.deserialize(buffer = t
 
 fun ByteArrayBuffer.hasRemaining(): Boolean = remaining > 0
 
+fun ByteArrayBuffer.plusOne(): ByteArrayBuffer {
```
zipdoki, line 23: @em3s 
Direct Kotlin port of `EdgeBuffer.plusOne` (codec-java); exposed as a `ByteArrayBuffer` extension so the cache prefix encoder doesn't need to wrap in an `EdgeBuffer`

em3s: Good 👍

em3s:
Nice Work! @zipdoki 

Optimistic merging. 

As discussed offline, follow-up PR to:
- Align `scan` and `seek` signatures.
- `seek` is backed by `multi-scan`, so the source parameter should be `List<Any>` to match

em3s:
Close #303

PR #320, opened 2026-05-12
Add explicit enable/disable HBase table and replication action endpoints

em3s:
## Summary

Add four explicit action endpoints on the HBase datastore controller for the safe table-cleanup runbook: flip `replicationScope` first (online, while the table still serves writes), then disable, then drop. Splitting these out of `updateTable` (PUT) lets the replication flip run without the storage-in-use guard, which the runbook requires.

## Changes

- `POST .../tables/{name}/enable` and `.../disable` — keep the storage-in-use guard, mirror `updateTable`
- `POST .../tables/{name}/replication/enable` and `.../disable` — call `AsyncAdmin.modifyColumnFamily` on the existing CF; skip the in-use guard; idempotent when already at target scope
- All four require the production-admin role; `updateTable` (PUT) is unchanged
- Assumes a single column family (matches existing `HBaseAdmin` / `HBaseTableSchema`)

## How to Test

- `./gradlew :engine:test --tests "*.HBaseAdminTest"` — 10/10 pass (3 new replication tests)
- `./gradlew :server:build -x test` and `:server:spotlessCheck :engine:spotlessCheck` — pass
- Manual: run `replication/disable` against a real cluster while the table is enabled and serving writes; then full runbook (replication/disable → disable → delete on both clusters)

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/datastore/hbase/admin/HBaseAdmin.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/datastore/hbase/admin/HBaseAdmin.kt
@@ -3,6 +3,7 @@ package com.kakao.actionbase.engine.datastore.hbase.admin
 import com.kakao.actionbase.v2.engine.util.getLogger
 
 import org.apache.hadoop.hbase.ClusterMetrics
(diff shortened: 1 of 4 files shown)


em3s:
Merging

PR #324, opened 2026-05-13
Preserve next-page cursor on scan with post-filter

### LINKED ISSUE #323: Scan with `filters` drops next-page cursor
## Background

A scan with `filters` (post-filter) always returns `offset: null` / `hasNext: false`, breaking pagination. `ranges` (pre-filter) scans are unaffected.

Cause: `DataFrame.where()` (`engine/.../v2/engine/sql/DataFrame.kt:106-114`) builds a new `DataFrame` and drops `offsets` / `hasNext` / `stats`. The upstream HBase scan sets them correctly, but `applyPredicates` in `V2BackedTableBinding.scan()` erases them before `toV3()` reads the cursor, leaving it `null`.

The original design assumed `offset` mattered only for `ranges`, not `filters`. That conflates filter mode with the scan cursor — `offset` is a property of the underlying paginated scan, and a post-filter must be transparent to scan metadata. An empty page with `hasNext = true` is valid: the cursor reflects "is the underlying scan exhausted," not "how many rows passed the filter."

`filters` was rarely used in production. The bug surfaced when a service needed pagination + filter together and the only workaround was to drop `filters` and filter on the client side — effectively making `filters` unusable as a paginated API.

## Task

- [ ] Preserve `offsets` / `hasNext` / `stats` in `DataFrame.where()`
- [ ] Add an `offset` + `filters` regression test in `IndexFilterSpec`

## Done When

- A scan with `filters` returns a working next-page cursor
- The regression test verifies that two consecutive pages cover all matching rows with no duplicates or omissions

em3s:
## Summary

`DataFrame.where()` was constructing a new instance and dropping `offsets` / `hasNext` / `stats`, so any scan invoked with a `filters` (post-filter) argument returned `offset: null` / `hasNext: false`, breaking pagination.

`offset` is a property of the underlying paginated scan, not of the filter mode. A post-filter must be transparent to scan metadata: even if every row on a page is rejected by the filter, the cursor must survive as long as the underlying scan has more rows to read.

Closes #323

## Changes

- `DataFrame.where()` now returns `copy(rows = filteredRows)`, preserving `offsets` / `hasNext` / `stats` instead of dropping them.
- Added `scan with offset and filters` regression test in `QueryServiceSpec` (not `IndexFilterSpec` as originally written in the issue — that spec only covers the `ranges` channel via `otherPredicates`, not the `filters` post-filter channel).

## How to Test

```
./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.QueryServiceSpec"
```

The regression test paginates with `limit=2 + filters=permission:eq:me` and verifies that page 1 returns a non-null cursor and that page 2, resuming via that cursor, yields the remaining matching row.

## AI Assistance

  - Tool / model: claude code (opus 4.7)

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/sql/DataFrame.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/sql/DataFrame.kt
@@ -110,7 +110,10 @@ data class DataFrame(
                     predicate.eval(row, schema)
                 }
             }
-        return DataFrame(filteredRows, schema)
+        // Preserve offsets/hasNext/stats: the post-filter is transparent
+        // to the underlying paginated scan, whose cursor must survive
+        // even when zero rows pass the predicate on this page.
+        return copy(rows = filteredRows)
     }
 
     companion object {
--- a/engine/src/test/kotlin/com/kakao/actionbase/v2/engine/v3/QueryServiceSpec.kt
+++ b/engine/src/test/kotlin/com/kakao/actionbase/v2/engine/v3/QueryServiceSpec.kt
@@ -129,6 +129,40 @@ class QueryServiceSpec :
                 }.verifyComplete()
         }
 
+        "scan with offset and filters" {
+            // Regression: post-filter (`filters`) must not strip the next-page cursor.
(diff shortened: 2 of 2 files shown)


em3s:
Optimistic Merging

PR #336, opened 2026-05-15
Preserve SYNC response contract under systemMutationMode=ASYNC

em3s:
## Summary

When `systemMutationMode=ASYNC` overrides a table whose own mode is `SYNC`, derive the response status from the request's `EdgeOperation` (`INSERT → CREATED`, `UPDATE → UPDATED`, `DELETE → DELETED`) instead of returning `QUEUED`. Tables whose own mode is `ASYNC` keep returning `QUEUED`. Mutations are still queued via the WAL — only the response status changes.

Closes #335

## Changes

- Derive operation-based status in `MutationService` for `system=ASYNC + label=SYNC`
- Update `MutationServiceSystemAsyncSpec` — expect operation-based status for SYNC tables, add UPDATE/DELETE coverage
- Update `EdgeMutationForceSyncTest` — adjust the SYNC-table queue expectation

## How to Test

- `./gradlew :engine:test --tests "*MutationServiceSystemAsyncSpec"`
- `./gradlew :server:test --tests "*EdgeMutationForceSyncTest"`
- `./gradlew build`

## AI Assistance

- [x] This PR was written largely with AI assistance.
  - Tool / model: claude code (opus 4.7)

PR #340, opened 2026-05-16
Apply SYNC response contract preservation to the v2 mutation path

### LINKED ISSUE #339: Apply SYNC response contract preservation to the v2 mutation path
## Background

#336 only addressed the v3 mutation path. The v2 DML path is still in use in production, so the same fix needs to be applied there.

## Task

- [ ] Apply the rule from #336 to the v2 mutation path (`v2.engine.Graph`):
  - `system=ASYNC + table=SYNC + !force` → `INSERT→CREATED` / `UPDATE→UPDATED` / `DELETE→DELETED`
  - other paths keep returning `QUEUED`
- [ ] Add v2 tests mirroring the v3 coverage

## Done When

- v2 behaves the same as v3 under `systemMutationMode=ASYNC`
- Tests cover all mode combinations × INSERT / UPDATE / DELETE

## Notes

Backport of #336. Same trade-off as #335.

em3s:
## Summary

Backport #336 to the v2 mutation path. When `systemMutationMode=ASYNC` overrides a table whose own mode is `SYNC`, derive the response status from the request's `EdgeOperation` (`INSERT → CREATED`, `UPDATE → UPDATED`, `DELETE → DELETED`, `PURGE → PURGED`) instead of returning `QUEUED`. Tables whose own mode is `ASYNC` keep returning `QUEUED`. Mutations are still queued via the WAL — only the response status changes.

Closes #339

## Changes

- Derive operation-based status in `v2.engine.Graph.mutate` for `system=ASYNC + table=SYNC` (no force)
- Add `v2.engine.dml.GraphSystemAsyncSpec` mirroring the v3 coverage — INSERT / UPDATE / DELETE / PURGE on SYNC EDGE, INSERT on SYNC MULTI_EDGE, ASYNC table keeps `QUEUED`, request=SYNC and force=true overrides

## How to Test

- `./gradlew :engine:test --tests "*v2*SystemAsync*"`
- `./gradlew build`

## AI Assistance

  - Tool / model: claude code (opus 4.7)


em3s:
Skipped `MULTI_EDGE` and multi-event tie-break coverage from #336: `MULTI_EDGE` is v3-only (v2 has compat but no integrations) and v2 `mutate()` takes a single operation per call, so multi-event cases cannot arise. Rationale captured in `GraphSystemAsyncSpec` KDoc.

em3s:
Optimistic merging.

Issue #322, opened 2026-05-13
Pruner for multi-hop EdgeCache (fat row prevention)


em3s:
## Background

Split from #214 (Phase 4). The multi-hop query path introduced a wide
row structure (EdgeCache) to serve 2nd-hop reads via multi-get. Without
a pruning mechanism, these wide rows can grow unbounded and become
fat, degrading read performance.

## Task

Two PRs, as agreed offline in #214:

- [ ] PR 1: Pruner API implementation
- [ ] PR 2: Pipeline code — extracts pruning candidates via WAL/CDC
      and calls the Pruner API

## Done When

- Pruner API enforces per-row size limits and evicts entries
- Pipeline processes WAL/CDC events and triggers pruning
- E2E coverage for the prune path

## References

- Parent: #214
- Phase 0 (skeleton): #217
- Phase 1 (mutation + multi-get): #226
- Phase 2/3 (n-hop + server): #243

zipdoki:
@em3s 
We had an offline meeting to discuss the EdgeCache prune API design. Here's a summary of what we decided.

---

## prune API skeleton

```
POST /graph/v3/databases/{database}/tables/{table}/edges/prune
```

- **Request**: `targets` — a list of `(start, direction)` pairs
- **Response**: `results` — a list of `(start, direction, cache, status)`

```
$ curl -X POST "https://.../graph/v3/databases/mydb/tables/mytable/edges/prune" \
  -H "Content-Type: application/json" \
  -d '{
    "type": "cache",
    "targets": [
      { "start": "user:1001", "direction": "OUT" }
    ]
  }'

{
  "results": [
    {
      "start": "user:1001",
      "direction": "OUT",
      "cache": "permission_created_at_desc",
      "status": "PRUNED"
    }
  ]
}
```

## Race condition handling
 prune is intentionally non-atomic, race conditions can occur between the qualifier scan and the subsequent delete.
 
| Strategy | Retained up to | Insert | Delete | Update (delete → insert) |
|---|---|---|---|---|
| limit only | ~100 | ✅ | ❌ | ❌ |
| limit + tolerance | ~200 | ✅ | ⚠️ | ⚠️ |
| unbounded | ∞ | ✅ | ✅ | ✅ |

(✅ consistent / ⚠️ may be inconsistent under high concurrent deletes / ❌ inconsistent)

- We decided to retain up to **`limit + tolerance `**:
  - `limit` — guarantees what readers see
[...]

PR #375, opened 2026-06-06
Add tolerance to cache metadata

zipdoki:
## Summary

Adds a `tolerance` field to cache metadata, bounding the margin of concurrent delete/update operations absorbed during cache prune (on top of `limit`).

When absent, `tolerance` defaults to `limit * 2` so older metadata deserializes with a concrete value. An explicit value is kept as-is; an explicit `null` is rejected.

Follows up on #322.

## Changes

- Add `tolerance: Int` to `Cache`, defaulting to `limit * 2`, validated as `> 0`
- Add deserialization and V2 ↔ V3 conversion tests

## How to Test
`./gradlew :core:test :server:test`

## AI Assistance

- [x] This PR was written largely with AI assistance

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/Cache.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/Cache.kt
@@ -9,10 +9,12 @@ data class Cache(
     val cache: String,
     val fields: List<CacheField>,
     val limit: Int = 100,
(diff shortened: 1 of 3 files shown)


em3s:
Could you add a comment explaining what limit and tolerance mean here? Also, please clarify how many items will be retained.

@zipdoki

em3s:
stale

PR #379, opened 2026-06-11
Validate edge schema before bulk encoding

### LINKED ISSUE #377: Add schema validation before HFile generation in bulkload (codec-java)
## Background

During a bulkload, edges are encoded into HFiles by `codec-java` and loaded
directly into HBase. There is currently **no schema validation before HFile
generation**, so an edge that violates the schema (e.g. a `nullable=false`
field set to `null`) can be written into HFiles and served as active data.

This surfaced in a staging bulkload: an edge with a required property set to
`null` was loaded without validation. It either failed the entire pipeline
(when an `UPDATE` replay for the same edge happened to fail) or — absent such a
replay — would have been served as active, schema-invalid data, confirmed via a
direct `GET` returning a required property as `null`.

**Root cause in codec-java**: `Edge.ensureType(EdgeSchema)` casts property
values via `DataType.cast()`, which silently returns `null` on a null/failed
cast (note the `// This allows null values` comment) and never consults
`Field.isNullable()`. Required fields that are null, mistyped, or missing from
the props map pass through unvalidated.

## Task

- [ ] Add a schema validation step in `codec-java` before HFile serialization that rejects schema-violating edges.
- [ ] Validate that every `nullable=false` field is present and non-null after casting.
- [ ] Also reject required fields that are entirely **absent** from the props map (not only present-but-null).
- [ ] Surface a clear, actionable error (offending field name + edge identity) instead of a silent pass or a generic NPE.
- [ ] Add tests: required field null / missing / type-mismatch, plus the valid happy-path case.

## Done When

- An edge with a `nullable=false` property that is null, missing, or mistyped is rejected **before** HFile generation with a descriptive error.
- Valid edges encode unchanged (no regression in existing `BulkEdgeEncoder` tests).
- Failure-case and happy-path tests are added and pass.

## Notes

- Likely change points: `Edge.ensureType()` (`com.kakao.actionbase.v2.core.edge.Edge`) and/or `BulkEdgeEncoder.bulkEncodeAll()`.
- `Field` already exposes `isNullable()`; `ensureType()` simply doesn't consult it today.
- Relevant tests: `BulkEdgeEncoderTests`, `MultiEdgeBulkEdgeEncoderTests`, `EdgeSerDerTest`.

eazyhozy:
## Summary

Add schema validation in `codec-java` before bulk encoding. Edges that violate the schema are now rejected with a descriptive error instead of being silently written as invalid active data.

Closes #377

## Changes

- **`Edge.validate(EdgeSchema)`** (new method): iterates all declared schema fields and rejects:
  - required (`nullable=false`) fields that are missing or null
  - any field with a non-null value that cannot be cast to its declared type (prevents the silent `null` coercion in `DataType.cast()`)
- **`BulkEdgeEncoder.bulkEncodeAll`**: calls `validate` for active edges immediately after `ensureType` and before any byte encoding. Inactive (tombstone) edges skip validation since property values are irrelevant for key-based deletion.
- **`EdgeValidationTest`** (new): unit tests covering required-missing, required-null, required-type-mismatch, nullable-type-mismatch, and happy-path cases.
- **`BulkEdgeEncoderTests`**: two integration tests verifying the bulk path rejects required-null and required-missing edges end-to-end.

## How to Test

```
./gradlew :codec-java:test
```

All existing tests continue to pass (no regression). New tests cover the failure cases from #377.

## AI Assistance

  - Tool / model: Claude Code / claude-opus-4-8

### DIFF
--- a/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/BulkEdgeEncoder.java
+++ b/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/BulkEdgeEncoder.java
@@ -63,6 +63,9 @@ public class BulkEdgeEncoder {
     int labelId = label.getId();
     Active active = bulkLoadEdge.isActive() ? Active.ACTIVE : Active.INACTIVE;
     Edge castedEdge = bulkLoadEdge.ensureType(label.getSchema());
+    if (bulkLoadEdge.isActive()) {
+      bulkLoadEdge.validateAgainstSchema(label.getSchema());
+    }
 
     List<TypedKeyFieldValue<T>> edges = new ArrayList<>();
 
--- a/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/BulkEdgeEncoder.java
+++ b/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/BulkEdgeEncoder.java
@@ -64,7 +64,7 @@ public class BulkEdgeEncoder {
     Active active = bulkLoadEdge.isActive() ? Active.ACTIVE : Active.INACTIVE;
     Edge castedEdge = bulkLoadEdge.ensureType(label.getSchema());
     if (bulkLoadEdge.isActive()) {
-      bulkLoadEdge.validateAgainstSchema(label.getSchema());
+      bulkLoadEdge.validate(label.getSchema());
     }
 
     List<TypedKeyFieldValue<T>> edges = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


eazyhozy:
Follow-up: the same silent null coercion affects the DML write path (nullable field type mismatch in `AbstractLabel.mutate`). Tracked in #380.

em3s:
Post-review refactoring (09c8e09):

- **Rename**: `Edge.validateAgainstSchema(EdgeSchema)` → `Edge.validate(EdgeSchema)`. The argument type already conveys what is validated against, so the redundant suffix was dropped. Matches the style of the neighboring `ensureType(schema)`.
- **Remove `continue`**: replaced the early-continue guard inside the loop with an `if / else if` block, making the two branches structurally parallel and easier to read.
- **PR title/body**: removed "HFile" references — `codec-java` only produces byte-encoded key/value pairs; the HFile write happens in the caller. Updated to "Validate edge schema before bulk encoding" to match the actual scope and the imperative style of other PRs in this repo.

PR #381, opened 2026-06-14
Fix MultiEdge group counter inconsistency on _source change

zipdoki:
## Summary

When a MultiEdge record's `_source` is updated, the old implementation submitted a decrement on the old source row and an increment on the new source row in a single batchAll call. Since HBase does not guarantee atomicity across rows, a partial failure could corrupt group counters with no safe retry path — replaying the same UPDATE would double-decrement the old source.

This PR fixes this by splitting the UPDATE into two sequential batches in V2BackedTableBinding.write:
1. Phase 1 (DELETE): persists an intermediate state (_source=<new>, active=false) and cleans up old source records
2. Phase 2 (CREATE): persists _source=<new>, active=true and populates new source records
  
The intermediate state written in phase 1 is the key to safe retries — if phase 2 fails and the client retries, read() picks up _source=<new>, active=false and re-enters only the CREATED path, preventing double-decrement on the old source counter.

## Changes

- V2BackedTableBinding.write: detect _source change on a MultiEdge UPDATE and execute DELETE + CREATE as two sequential handleDeferredRequests calls instead of one  

## How to Test
  - `./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.edge.MultiEdgeSpec"`

## AI Assistance

  - Tool / model:  Claude Sonnet 4.6 (claude.ai/code)

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedTableBinding.kt
@@ -104,10 +104,12 @@ class V2BackedTableBinding(
         val beforeRecord = EdgeStateRecord.of(source, target, beforeClean, label.entity.id)
         val afterRecord = EdgeStateRecord.of(source, target, afterClean, label.entity.id)
 
-        val isMultiEdgeSourceChanged = descriptor.schema is ModelSchema.MultiEdge &&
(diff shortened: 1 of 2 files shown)


zipdoki:
Closing this PR — the proposed fix splits the mutation into two sequential batchAll calls, but each batch still contains cross-row operations internally. The partial failure risk is identical to the original implementation, so this change provides no meaningful improvement. A different approach is needed to address this properly.

Issue #369, opened 2026-06-01
Support universal Top-K query engine


em3s:
## Background

Add a universal Top-K query engine to support precomputed ranking queries (e.g., trending lists, leaderboards, user preferences) as first-class operations.

Conceptually, `topk` acts as a ranking operator over Actionbase's aggregation engine (`agg`):
```
topk(agg(...))
```

Consequently, **all aggregation logic (e.g., counts, sums, weighted multi-edge merging, time-decay) must reside within the `agg` layer.** The `topk` engine is strictly responsible for performing flat O(K) ranking queries over these pre-aggregated states, ensuring a clean decoupling of write-time aggregation and read-time ranking.

### Interface

```
topk(Target | [Filter], [Window])
```

The engine will support **4 query shapes**:

| Shape | Signature | Example |
|---|---|---|
| 1 | `topk(Target)` | Most active users (All time) |
| 2 | `topk(Target \| Window)` | Trending items this week |
| 3 | `topk(Target \| Filter)` | Most purchased items by a user |
| 4 | `topk(Target \| Filter, Window)` | Recent category heavy users / User's recent purchases |

## Done When

- [ ] Universal `topk` interface is supported, and **"top purchased items per user within the last year (Shape 4)"** works as the initial target.
- [ ] Read latency remains bounded at O(K), independent of total interaction volume.
- [ ] Comprehensive unit/integration tests and documentation are completed.

## Notes

- **Decoupled Writes**: Index updates for ranking must not block or bottleneck the primary write/mutation path.
- **Pre-aggregated State**: Top-K lists are maintained via bucket pre-aggregation to avoid full-scan lookups.
- **Signature Degradation**: Since Shape 4 is the superset query signature, implementing it with optional `Filter` and `Window` parameters naturally satisfies Shapes 1, 2, and 3 as omitted parameter fallbacks.

em3s:
Partially resolved by #500. I'll pick up the rest later.

PR #387, opened 2026-06-30
Support `__count__` sentinel in AGG view to delegate to native COUNT query

eazyhozy:
## Summary

Add `__count__` as a reserved sentinel group name in the AGG endpoint (`GET .../edges/agg/{group}`). When the group is `__count__`, the request is internally delegated to the existing native edge-count query instead of looking up a pre-defined aggregation group in table metadata.

Part of #369

## Changes

- `core/.../Constants.kt` — add `Constants.Group.COUNT_SENTINEL = "__count__"` alongside the existing reserved-marker convention (`VERTEX_MARKER`)
- `engine/.../service/QueryService.kt` — detect `__count__` in `agg()`, delegate to `counts()` (native HBase counter path), and map `DataFrameEdgeCountPayload` → `DataFrameEdgeAggPayload` (`count` field → `value` field). Add `require(ranges == null)` guard for the sentinel path. Make `ranges` nullable; non-sentinel groups enforce non-null via `requireNotNull`.
- `server/.../EdgeQueryController.kt` — make `@RequestParam ranges` nullable (default `null`) so callers can omit it when using `__count__`
- `server/.../EdgeAggCountSentinelE2ETest.kt` — 5 E2E test cases (first E2E coverage for the `/edges/agg/{group}` endpoint):
  - OUT/IN count: asserts `start`, `direction`, and `value` fields; `value` is compared against `/edges/count` as ground truth so the two paths are verified to be consistent
  - Multi-start: asserts per-node `value` for each start key
  - `ranges` + `__count__` → 400
  - Real schema group (`by_category`) without `ranges` → 400 (exercises the `requireNotNull(ranges)` path)

`TableBinding` interface is unchanged — all binding implementations benefit automatically.

## How to Test

```bash
./gradlew :server:test --tests "*.EdgeAggCountSentinelE2ETest"
```

Or end-to-end against a running server:

```bash
# Create a table, insert some edges, then:
curl ".../edges/agg/__count__?start=1&direction=OUT"
# → { "groups": [{ "start": 1, "direction": "OUT", "value": <N>, "context": {} }], "count": 1, "context": {} }

# ranges must be omitted for __count__
curl ".../edges/agg/__count__?start=1&direction=OUT&ranges=foo%3Dbar"
# → 400 Bad Request
```

## AI Assistance

  - Tool / model: Claude Code (claude-sonnet-4-6)

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
@@ -35,5 +35,8 @@ object Constants {
 
     object Group {
         const val DEFAULT_TTL = 8 * 24 * 60 * 60 * 1000L // 8 days in milliseconds (691,200,000)
+
+        /** Sentinel group name that delegates to the native edge-count query. */
+        const val COUNT_SENTINEL = "__count__"
     }
 }
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/service/QueryService.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/service/QueryService.kt
@@ -1,10 +1,12 @@
 package com.kakao.actionbase.engine.service
 
+import com.kakao.actionbase.core.Constants
 import com.kakao.actionbase.core.Constants.VERTEX_MARKER
 import com.kakao.actionbase.core.edge.EdgeField
 import com.kakao.actionbase.core.edge.payload.DataFrameEdgeAggPayload
 import com.kakao.actionbase.core.edge.payload.DataFrameEdgeCountPayload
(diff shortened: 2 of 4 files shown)


eazyhozy:
@em3s Could you take a look when you have a moment?

This PR adds `__count__` as a sentinel group name in the AGG endpoint — when `group == "__count__"`, the request is internally delegated to the native edge-count query instead of resolving a metadata-defined group. Part of the unified TopK work (#369).

em3s:
LGTM.

Nit: consider rejecting group names that start with _ (or the sentinel itself) at schema creation time.

Nice work @eazyhozy !

PR #391, opened 2026-07-03
Add migrate CLI command with datastore:// storage rewrite (Phase 2, #347)

em3s:
## Summary

Adds a native `migrate` CLI command to seed metadata into HBase for Phase 2 of #347, replacing the earlier `dev/*.py` prototype scripts (never merged). Each label's storage reference is rewritten to a `datastore://` URI during migration, so the seeded metadata is valid against the new storage model (#398).

**Once migration completes, MySQL is effectively dead.** Every key exists in HBase, so the overlay (#390) always returns the HBase value and never falls through to MySQL. No code change needed — MySQL just stops being read. That's the gate for Phase 3.

I rebuilt the branch on current main: it previously stacked on #390's overlay commits, which have since been squash-merged. Re-verifying against main surfaced three contract drifts, fixed here: duplicates surface as `400` (not `409`), only HBASE storages have a `datastore://` equivalent, and label `caches` were dropped from the replay body.

Related: #347, #390, #398

## Changes

- `cli/internal/command/migrate.go` — new command, two phases:
  - `migrate plan [-o <file>]` reads databases/tables/aliases (sys/origin skipped) and writes a `{path, body, label}` plan
  - `migrate apply [-i <file>]` replays each entry; duplicates are skipped so re-runs are idempotent
- Duplicate detection matches the server's actual contract: the v2 DDL path maps `failOnExist` duplicates to `400` with an "already exists" message (`AbstractLabel` → `ExceptionController`) — there is no `409`. A plain `409` is still treated as a skip.
- Storage rewrite is HBASE-only: `HBASE → datastore://<namespace>/<table>`. The engine resolves every `datastore://` URI except the reserved `__sys__` metastore as an HBase namespace/table, so minting sentinel URIs for JDBC/LOCAL would create labels the server can't route. `migrate plan` fails fast, with the reason, if a label references a non-HBASE storage. References that are already `datastore://` URIs pass through unchanged.
- Label `caches` are carried into the replay body (previously dropped silently).
- `client.PostRaw` / `http.PostForResult` — POST returning status + raw body, so apply can key on the error message; failures print the server response
- `command_type.go`, `actionbase_runner.go` — register the command
- Unit tests for the storage→URI conversion, the `tableBody` rewrite paths, duplicate matching, and the sys/origin filter

## How to Test

```
cd cli && go test ./...
ACT_HOST=https://ab.example.com ACT_API_KEY=<key> actionbase -e "migrate plan"
# review migration-run.json, then
ACT_HOST=https://ab.example.com ACT_API_KEY=<key> actionbase -e "migrate apply"
```

`migrate apply` exits non-zero if any entry fails.

## AI Assistance

  - Tool / model: Claude Opus 4.8, Claude Fable 5

### DIFF
--- a/cli/internal/client/actionbase_client.go
+++ b/cli/internal/client/actionbase_client.go
@@ -111,6 +111,12 @@ func (a *ActionbaseClient) Mutate(
 	)
 }
 
+// PostRaw POSTs an arbitrary body to a path and returns the HTTP status code.
+// Used by migrate apply to replay a plan without knowing each entry's type.
+func (a *ActionbaseClient) PostRaw(path string, body any) int {
+	return PostForStatus(a.client, path, body)
+}
+
 func (a *ActionbaseClient) GetHost() string {
 	return a.client.baseUrl
 }
--- a/cli/internal/client/actionbase_client.go
+++ b/cli/internal/client/actionbase_client.go
@@ -111,10 +111,11 @@ func (a *ActionbaseClient) Mutate(
 	)
 }
 
-// PostRaw POSTs an arbitrary body to a path and returns the HTTP status code.
-// Used by migrate apply to replay a plan without knowing each entry's type.
-func (a *ActionbaseClient) PostRaw(path string, body any) int {
-	return PostForStatus(a.client, path, body)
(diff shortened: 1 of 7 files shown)


em3s:
Optimistic

Issue #389, opened 2026-07-02
Encode EdgeGroup records in the bulk-load path (BulkEdgeEncoder)


eazyhozy:
## Background

The bulk-load encoding path (`BulkEdgeEncoder.bulkEncodeAll` in `codec-java`) currently encodes only **EdgeState / EdgeIndex / EdgeCache / EdgeCount** from a `BulkLoadEdge`. It does **not** encode **EdgeGroup** (aggregation) records.

As a result, a table initialized via bulk-load has empty aggregation data until records accumulate incrementally through the online write path. This blocks onboarding new tables (and large migrations) that need aggregation available right after the initial load.

For contrast, the online path already builds group records in `EdgeMutationBuilder.buildGroupRecords()` (`core`) and writes them as HBase `Increment`s. `codec-java` has no EdgeGroup encoding at all — the `groups` metadata on a label is parsed (`LabelDTO.getGroups()`) but never consumed by any encoder.

## What this change does

`codec-java` will encode the **full EdgeGroup record bytes** for each edge:

- **row key**: `hash | directedSource | tableCode | EDGE_GROUP(-5) | direction.code | groupCode`
- **qualifier**: `groupValues` encoded with `Order.DESC` (including bucketed field values)
- **value**: this single edge's contribution — `1` for COUNT, the `valueField` value for SUM

Because many edges can contribute to the same `(row key, qualifier)`, summing those contributions into the final absolute cell value is done downstream by the external bulk-load pipeline (group-by + sum, then written to HFiles). This is exactly how `EdgeCount` already works: `codec-java` encodes the counter key per edge, and the pipeline aggregates the counts. The encoding is fully owned by `codec-java`; only the cross-edge aggregation is out of scope here.

## Design constraints

- **EdgeCache is the precedent.** EdgeCache encoding was added to `BulkEdgeEncoder` first, ahead of the external pipeline consuming it. EdgeGroup follows the same two-step approach: land the codec encoding here first, then update the pipeline separately.
- **Record-type byte `-5` collides.** `EncodedEdgeType.IMMUTABLE_INDEXED_EDGE_TYPE` (`-5`, dead code — "Currently not used anywhere") occupies the same physical record-type tag as the online `EDGE_GROUP` record type (`-5`). We should reuse `-5` as `EDGE_GROUP` in `codec-java`.
[...]

PR #393, opened 2026-07-07
Free up EDGE_TYPE=-5 for EDGE_GROUP_TYPE

eazyhozy:
## Summary

Splits off the `-5` rename tweak from #389, per @zipdoki's review comment.

`EncodedEdgeType.IMMUTABLE_INDEXED_EDGE_TYPE` occupied `-5` but was never wired to any live encoder/decoder path. Renaming it in place to `EDGE_GROUP_TYPE` would have implied `-5` had always meant EDGE_GROUP, which isn't historically true. Instead:

- Move `IMMUTABLE_INDEXED_EDGE_TYPE` to an unused code (`-100`).
- Add `EDGE_GROUP_TYPE(-5)` as a new entry, matching core's `EdgeRecordType.EDGE_GROUP(-5)`.
- Clean up the dead `-5` references in `AbstractEdgeEncoder` / `DecodedEdge`.

No behavior change.

Ref #389

## Changes

- `EncodedEdgeType.java`: move `IMMUTABLE_INDEXED_EDGE_TYPE` to `-100`, add `EDGE_GROUP_TYPE(-5)`.
- `AbstractEdgeEncoder.java`: drop the dead `IMMUTABLE_INDEXED_EDGE_TYPE` constant (no callers referenced it).
- `DecodedEdge.java`: drop the dead `IMMUTABLE_INDEXED_EDGE_TYPE` decode branches (L98-99, L164).
  - This type code was never actually produced on the wire — `BulkEdgeEncoder` encodes `LabelType.IMMUTABLE_INDEXED` labels using the `INDEXED_EDGE_TYPE` tag, not `IMMUTABLE_INDEXED_EDGE_TYPE`, so no encoded byte stream ever carried this type code and the branch was unreachable.
  - Since `IMMUTABLE_INDEXED_EDGE_TYPE` is being moved off `-5` anyway, leaving the branch in place (just comparing against `-100` instead) would keep dead code around under a new code point rather than actually cleaning it up as the review comment asked.
  - Verified via `grep -rn "IMMUTABLE_INDEXED_EDGE_TYPE"` across `codec-java`: the only references were this decode branch and the unused `AbstractEdgeEncoder` constant.

## How to Test

```
./gradlew :codec-java:test
```

All existing tests pass; no behavior change.

## AI Assistance

  - Tool / model: Claude Code (Sonnet 5)

### DIFF
--- a/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/AbstractEdgeEncoder.java
+++ b/codec-java/src/main/java/com/kakao/actionbase/v2/core/code/AbstractEdgeEncoder.java
@@ -22,8 +22,6 @@ public abstract class AbstractEdgeEncoder<T> implements EdgeEncoder<T> {
   public static final byte COUNTER_EDGE_TYPE = EncodedEdgeType.COUNTER_EDGE_TYPE.getCode();
 
   public static final byte INDEXED_EDGE_TYPE = EncodedEdgeType.INDEXED_EDGE_TYPE.getCode();
-  public static final byte IMMUTABLE_INDEXED_EDGE_TYPE =
-      EncodedEdgeType.IMMUTABLE_INDEXED_EDGE_TYPE.getCode();
   public static final byte EDGE_CACHE_TYPE = EncodedEdgeType.EDGE_CACHE_TYPE.getCode();
 
   public static final String INSERT_TS_KEY = "__InsertTs__";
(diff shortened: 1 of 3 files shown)


zipdoki (approved):
LGTM 👍

eazyhozy:
Offline review feedback from @em3s:

> Since `IMMUTABLE_INDEXED_EDGE_TYPE` is being removed from the branching logic anyway, it's an enum entry that doesn't seem likely to be used again — better to just remove it entirely.

Agreed — will remove `IMMUTABLE_INDEXED_EDGE_TYPE` entirely instead of moving it to `-100`.

zipdoki:
Optimistic Merge

PR #395, opened 2026-07-07
Fix AGG on MULTI_EDGE _source/_target group fields

### LINKED ISSUE #394: Agg query returns empty for MULTI_EDGE groups on _source/_target
## Background

MULTI_EDGE tables promote the top-level source / target fields into the persistence layer under _source / _target.

Today the AGG query returns an empty result. Writes are fine, but reads cast the range predicate through MultiEdge.propertiesByName, which only exposes user-declared properties. _source / _target fall through the cast, land as String, and the resulting qualifier bytes never match the Long written at insert time.

## Task

- [x] Resolve _source / _target to the top-level MultiEdge.source / target primitive types when casting AGG range predicates in V2BackedTableBinding.
- [x] Add a MULTI_EDGE + _target group case to EdgeAggQuerySpec.

## Done When

- AGG on a MULTI_EDGE group field of _source or _target returns the recorded counts.
- The new spec passes.
- No changes to ModelSchema public API, propertiesByName, or non-MULTI_EDGE behavior.

## Notes

- Follows #388, which introduced the type cast path in encodeAggRanges.
- The domain view (ModelSchema.MultiEdge) keeps source / target as top-level fields; only the persistence-layer keys are _source / _target. The read cast is aligned with the domain view rather than expanding properties.

zipdoki:
## Summary

Follow-up to #388. AGG on MULTI_EDGE group fields _source / _target returned empty: the cast site looked them up in propertiesByName, but those are persistence-layer keys not exposed on the domain view. Predicates fell back to String while writes stored Long, so qualifier bytes never matched.

Resolve _source / _target to MultiEdge.source.type / target.type at the cast site. ModelSchema is untouched.

Closes #394

## Changes

- Add ModelSchema.fieldType(name) helper in V2BackedTableBinding covering the promoted keys.
- Route both AGG cast sites (agg, bucketOrCast) through it.
- Add a MULTI_EDGE + _target group case to EdgeAggQuerySpec.

## How to Test
`./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.EdgeAggQuerySpec"`

## AI Assistance

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/ModelSchema.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/ModelSchema.kt
@@ -22,7 +22,7 @@ sealed class ModelSchema : AbstractSchema {
     abstract val properties: List<StructField>
 
     @get:JsonIgnore
-    val propertiesByName: Map<String, StructField> by lazy { properties.associateBy { it.name } }
+    open val propertiesByName: Map<String, StructField> by lazy { properties.associateBy { it.name } }
 
     @JsonTypeName("edge")
     data class Edge(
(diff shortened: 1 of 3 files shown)


em3s:
LGTM 👍 @zipdoki

PR #396, opened 2026-07-08
Skip __count__ sentinel group in mutation path

eazyhozy:
## Summary

`__count__` is a sentinel group name reserved for AGG query interface compatibility and global topk metadata, added in #387. On read, it delegates to the native EdgeCount query instead of a real EdgeGroup lookup, so it never needs an EdgeGroup row of its own. The mutation path didn't know about this yet: if a table's `groups` metadata included `__count__`, every edge insert/update/delete would still emit an unnecessary `EdgeGroupRecord` (HBase `Increment`) for it.

Follow-up to #387, which introduced the `__count__` sentinel on the read path.

## Changes

- `core/.../edge/mutation/EdgeMutationBuilder.kt` — `buildGroupRecords()` now skips any group whose name equals `Constants.Group.COUNT_SENTINEL`, before building its `EdgeGroupRecord`s. This function is the single choke point for all three mutation transitions (CREATED / DELETED / UPDATED), so one guard covers the whole write path.
- `core/.../edge/mutation/EdgeMutationBuilderTest.kt` — 3 new cases: CREATED with only a `__count__` group produces no `groupRecords` (Edge and MultiEdge strategies); a `__count__` group mixed with a real group is skipped while the real group's records are still produced correctly.

Out of scope: bulk-load doesn't have an EdgeGroup encoder yet (#389); the same skip will need to be added there once it exists.

## How to Test

```bash
./gradlew :core:test --tests '*EdgeMutationBuilderTest'
```

Regression check for the existing sentinel read path:

```bash
./gradlew :engine:test --tests '*EdgeAggQuerySpec'
./gradlew :server:test --tests '*EdgeAggCountSentinelE2ETest'
```

## AI Assistance

  - Tool / model: Claude Code (claude-opus-4-8)

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
@@ -36,22 +36,7 @@ object Constants {
     object Group {
         const val DEFAULT_TTL = 8 * 24 * 60 * 60 * 1000L // 8 days in milliseconds (691,200,000)
 
-        /**
-         * Group names reserved for special handling by the engine. An entry here is
-         * the single source of truth: the read (AGG) and mutation paths both derive
-         * their sentinel checks from this enum, so adding a new reserved name never
-         * requires touching more than one place.
-         */
-        enum class Reserved(val groupName: String) {
-            /** Delegates to the native edge-count query instead of a real EdgeGroup lookup. */
-            COUNT_SENTINEL("__count__"),
-            ;
-
-            companion object {
-                private val names = entries.map { it.groupName }.toSet()
-
-                fun contains(groupName: String): Boolean = groupName in names
-            }
-        }
+        /** Sentinel group name that delegates to the native edge-count query. */
+        const val COUNT_SENTINEL = "__count__"
     }
 }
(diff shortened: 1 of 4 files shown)


eazyhozy:
Addressed offline review feedback from @em3s: don't special-case `COUNT_SENTINEL` alone in the mutation guard — use a reserved-name exclusion set so future sentinels don't require touching multiple places.

Follow-up commit replaces the standalone `Constants.Group.COUNT_SENTINEL` constant with `Constants.Group.Reserved`, an enum of reserved group names. `EdgeMutationBuilder.buildGroupRecords()` now checks membership via `Reserved.contains(group.group)` instead of comparing against a single constant, so adding a new reserved name is a one-line enum entry with no separate set to keep in sync.

eazyhozy:
Follow-up on the exclude-set approach: I initially replaced the constant with a `Reserved` enum so future sentinels couldn't be added to `Constants.Group` without also declaring whether they skip group mutation. On reflection this was more machinery than the read path uses for the same sentinel — `QueryService.agg` still does a plain `group == Constants.Group.COUNT_SENTINEL` check, no enum involved. Reverted: `Constants.Group.COUNT_SENTINEL` stays a single constant, and the exclude set now lives in `EdgeMutationBuilder` (`groupNamesExcludedFromMutation`), next to the mutation-skip behavior it actually drives. Same effect as requested — new reserved names get added to that one set — without introducing a second mechanism.

zipdoki:
Optimistic Merge

PR #397, opened 2026-07-08
Encode EdgeGroup records in the bulk-load path (BulkEdgeEncoder)

### LINKED ISSUE #389: Encode EdgeGroup records in the bulk-load path (BulkEdgeEncoder)

eazyhozy:
## Summary

The bulk-load encoding path (`BulkEdgeEncoder.bulkEncodeAll` in `codec-java`) only encoded EdgeState/EdgeIndex/EdgeCache/EdgeCount from a `BulkLoadEdge`. It never encoded EdgeGroup (aggregation) records, so tables initialized via bulk-load had empty aggregation data until it accumulated incrementally through the online write path.

This PR ports the online (V3) group-encoding model into `codec-java` (V2) and wires it into `BulkEdgeEncoder`, so bulk-loaded edges emit EdgeGroup records too.

Closes #389

## Changes

- `metadata/Bucket.java` (new): ports V3's `date` bucket — epoch value → formatted string — used to bucket group fields (e.g. daily aggregation). Query-time `handleQueryValue` is intentionally not ported; bulk encoding only ever applies buckets to concrete property values.
- `metadata/Group.java`: `Group.Field` gains an optional `bucket` and an `applyBucket` helper.
- `code/EdgeEncoder.java` / `code/AbstractEdgeEncoder.java` / `code/BytesKeyValueEdgeEncoder.java` / `code/StringKeyFieldValueEdgeEncoder.java`: new `encodeGroupEdge`/`encodeAllGroupEdges` methods mirroring the existing EdgeCache encoding path. Encodes the full V3-compatible EdgeGroup record bytes for a single edge — row key (`hash|directedSrc|labelId|EDGE_GROUP|direction|groupCode`), qualifier (bucketed `groupValues`, `Order.DESC`, matching `EdgeGroupRecordMapper`'s `GROUP_RECORD_ORDER`), and this edge's single contribution as a raw 8-byte long (`1` for COUNT, the `valueField`'s value for SUM) — matching the raw `buffer.long` V3 uses for its HBase `Increment` cell, not an `OrderedBytes` header.
- `code/BulkEdgeEncoder.java`: wires EdgeGroup encoding into `bulkEncodeAll` for INDEXED and MULTI_EDGE labels, one row per `(group, direction)`, iterating each group's own `directionType` (not the label's) — matching V3 `EdgeMutationBuilder.buildGroupRecords`. Skips the `__count__` sentinel group, matching core's online mutation path (#396) — it delegates to the native EdgeCount query and never persists an EdgeGroup row.
- Tests: `BulkEdgeEncoderTests`, `MultiEdgeBulkEdgeEncoderTests`, `BucketTests` (codec-java), and `V2MultiEdgeBulkLoadTest` (core) — the bulk-encoded group key/qualifier/value round-trips through V3's `EdgeGroupRecordMapper` decoder byte-for-byte, the same safety net already used for EdgeCount/EdgeCache.

## Out of scope

Consuming these `EDGE_GROUP_TYPE` records downstream (aggregating per-edge contributions into absolute cell values, writing to HFiles) happens in the external bulk-load pipeline, not in this repo. This follows the same two-step precedent as EdgeCache: land the codec encoding first, update the pipeline separately.

## How to Test

```bash
./gradlew :codec-java:test
./gradlew :core:test --tests "*V2MultiEdgeBulkLoadTest*"
```

All tests pass, including the existing count/cache round-trips and the groups-absent backward-compatibility case.

## AI Assistance

  - Tool / model: Claude Code (Sonnet 5)


codec-java/src/main/java/com/kakao/actionbase/v2/core/code/AbstractEdgeEncoder.java
```diff
@@ -23,6 +28,7 @@ public abstract class AbstractEdgeEncoder<T> implements EdgeEncoder<T> {
 
   public static final byte INDEXED_EDGE_TYPE = EncodedEdgeType.INDEXED_EDGE_TYPE.getCode();
   public static final byte EDGE_CACHE_TYPE = EncodedEdgeType.EDGE_CACHE_TYPE.getCode();
+  public static final byte EDGE_GROUP_TYPE = EncodedEdgeType.EDGE_GROUP_TYPE.getCode();
```
eazyhozy: self-review: preserve ordering by code

```suggestion
  public static final byte EDGE_GROUP_TYPE = EncodedEdgeType.EDGE_GROUP_TYPE.getCode();
  public static final byte EDGE_CACHE_TYPE = EncodedEdgeType.EDGE_CACHE_TYPE.getCode();
```

codec-java/src/main/java/com/kakao/actionbase/v2/core/code/AbstractEdgeEncoder.java
```diff
@@ -222,6 +228,72 @@ private static Object resolveFieldValue(
...
+
+    return groups.stream()
+        .filter(group -> !GROUP_COUNT_SENTINEL.equals(group.getGroup()))
+        .flatMap(
+            group ->
+                group.getDirectionType().getDirs().stream()
+                    .map(dir -> encodeGroupEdge(ts, src, tgt, props, dir, labelId, group)))
+        .collect(Collectors.toList());
+  }
+
```
eazyhozy, line 291: self-review: preserve ordering consistency of definitions

em3s:
Is this ready for review? @eazyhozy 

As with the previous `EdgeCache` work, please verify that EdgeGroup works correctly in the internal bulkload path and share the test results.

https://github.com/kakao/actionbase/pull/309#issuecomment-4405277556

eazyhozy:
@em3s To verify the bulk-load path end-to-end without relying on the internal pipeline, I'd like to add a test in `core` that simulates the pipeline's aggregation step: encode several edges via `BulkEdgeEncoder`, group-by (row key, qualifier) + sum their per-edge contributions, and assert the result matches what the online path (`EdgeMutationBuilder`) produces for the same edges.

This would sit alongside the existing `V2MultiEdgeBulkLoadTest` and `EdgeMutationBuilderTest` in `core`. Does this address the verification concern?

em3s:
That's a great approach and worth adding.

However, to be completely confident, we still need to see it work end-to-end. Please run the internal pipeline test and share the results as well.

@eazyhozy

eazyhozy:
@em3s Ran the end-to-end test on the internal pipeline — confirmed that EdgeGroup data is generated correctly through the bulk-load path.

eazyhozy:
@em3s Follow-up with the full-scale end-to-end result. The earlier verification I posted was from a run that turned out to be incomplete: the bulk pipeline's encode step was writing the per-edge EdgeGroup contributions to HFiles without the group-by-and-sum aggregation, so same-bucket cells collapsed instead of accumulating. That aggregation is the pipeline's responsibility by design (as covered in V2BulkLoadAggregationTest), and after adding it on the pipeline side (group by row key + qualifier, sum contributions), the full end-to-end run passed: aggregated counts are correct across both whole-period and time-bucketed groups, verified against sampled anchor edges including events from 8 years back. Removing the duplicate rows also cut the bulk-loaded HBase row count by ~63%. No codec-java changes were needed — the encoder contract here (one contribution per edge, COUNT=1 / SUM=valueField, matching the online mutation path) held up as-is.

em3s:
Good job! Noted that the internal tests have passed. @eazyhozy

Please merge after a final check. @zipdoki

zipdoki:
Optimistic merge.

PR #400, opened 2026-07-08
Add topk aggregation metadata and list endpoint

zipdoki:
## Summary

Introduce top-K aggregation metadata on `Group` and expose a read endpoint listing tables that declare one.

A `Group` can now carry an `Aggregations` block. Today only `topk` is defined; the schema is designed so future aggregation kinds (e.g. HLL) can land as sibling fields without restructuring the payload.

`GET /graph/v3/aggregations` returns a flat list of `(type, database, table)` tuples for every table that declares an aggregation, filterable by `?type=`. The aggregate/mutate flow that consumes this metadata is out of scope for this PR and will land separately.

### Metadata example

A group inside a label creation payload:

```json
{
  "groups": [
    {
      "group": "purchased_count",
      "type": "COUNT",
      "fields": [{"name": "_target"}],
      "directionType": "OUT",
      "aggregations": {
        "topk": [
          {
            "topk": "top_purchased",
            "entity": "_source",
            "ranges": "_target:eq:{_target}",
            "dimension": "_target",
            "refreshAfterMillis": 31536000000,
            "rank": "commerce.purchases__topk"
          }
        ]
      }
    }
  ]
}
```

### Endpoint example

`GET /graph/v3/aggregations`

```json
{
  "items": [
    { "type": "TOPK", "database": "commerce", "table": "purchases" }
  ]
}
```

`GET /graph/v3/aggregations?type=TOPK` filters to a single aggregation type.

## Test plan

- `./gradlew :core:test --tests 'com.kakao.actionbase.core.metadata.common.AggregationsTest'`
- `./gradlew :engine:test --tests 'com.kakao.actionbase.engine.service.AggregationServiceSpec'`
- `./gradlew :engine:test --tests 'com.kakao.actionbase.v2.engine.entity.LabelEntityFunctionsSpec'`
- `./gradlew :server:test --tests 'com.kakao.actionbase.server.api.graph.v3.metadata.MetadataAggControllerE2ETest'`

## AI Assistance


em3s:
As we discussed offline about the `expire` / `refresh` terminology:

`expire` implies the value is invalid at that point., but it's still valid, just due for a refresh. 

Could you rename these throughout the PR?

- `expireAfterMillis` → `refreshAfterMillis`
- the `expire` boolean → `refresh` (it means "needs to be refreshed")

Also, since we decided the refresh table will be a single system-wide table, TopkTable no longer needs the expire field.

@zipdoki

core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/Group.kt
```diff
@@ -27,3 +28,36 @@ data class Group(
...
+    val score: String,
+)
+
+enum class AggregationType {
+    TOPK {
+        override fun has(aggregations: Aggregations): Boolean = aggregations.topk.isNotEmpty()
+    }, ;
+
+    abstract fun has(aggregations: Aggregations): Boolean
+}
```
zipdoki, line 61: Might look like overkill for a single variant, but a second aggregation kind is planned in a follow-up. Keeping the extension surface to "one enum case + one matching field on `Aggregations`" — every consumer (`isEmpty`, `supports`, `LabelEntity.hasAggregation`, `toQualifiedAggregations`) routes through here and stays untouched when new kinds land.

zipdoki:
@em3s 
Thanks! Applied in 123210c:
  
- `expireAfterMillis` → `refreshAfterMillis`
- Dropped `TopkTable.expire`

The boolean `expire` was already removed earlier on this branch when the user/system aggregation paths were unified, so no rename needed there.

Ready for another look.

zipdoki:
Closing — re-cutting this into smaller PRs off `main`. I'll link them here.

PR #402, opened 2026-07-09
Bump default codec buffer size to 8 KiB

zipdoki:
## Summary

Bump the default codec buffer size from 5120 to 8192 bytes.

8 KiB is a common page/IO size and stays well below HBase's MOB threshold (~100 KiB), so encoded rows stay on the regular write path.

Ref: https://hbase.apache.org/docs/architecture/hbase-mob/

## Changes

- Set Constants.Codec.DEFAULT_BUFFER_SIZE to 8192
- Add a short comment noting the MOB threshold

## How to Test
`./gradlew :core:test --tests "com.kakao.actionbase.core.codec.*"`

## AI Assistance


eazyhozy (dismissed):
~LGTM👍🏻~

offline review: tests for having 8k buffer are required

em3s:
Please add tests for 8 KiB and over 8 KiB. @zipdoki

Currently, the value is maintained in two places due to the legacy implementation. It would be better to cover both cases in one test file so we do not miss either side.

emeth-kim (approved):

em3s:
Since the value is still maintained in two places due to the legacy implementation, testing it in `engine` makes sense. 

LGTM 👍  @zipdoki

PR #403, opened 2026-07-09
Enforce lowercase alphanumeric/underscore naming across all metadata APIs

em3s:
## Summary

Enforces a uniform naming policy — `^[a-z][a-z0-9_]{0,63}$` — across **all** Actionbase metadata create paths: V3 (database, table, alias) and V2 (service, label, alias, storage). Names containing hyphens or uppercase letters are rejected with 400 Bad Request on create. Read, update, and delete paths are unaffected. Note: creating a new table/alias under a legacy-named parent database is rejected, since the parent name is validated on create — intentional, as it also stops new resources accreting under non-conforming databases.

### Why this naming policy

Two constraints that apply simultaneously drove the choice of `^[a-z][a-z0-9_]{0,63}$`:

1. **URL path safety** — Names appear directly in HTTP path segments. Uppercase causes case-sensitivity issues across clients, proxies, and load balancers. Hyphens are technically URL-safe but create ambiguity with path separators in some routing layers.
2. **HBase naming constraints** — Names map directly to HBase table and namespace identifiers. Allowing names that pass HTTP validation but fail HBase validation produces confusing runtime errors deep in the stack.

The intersection of these two constraint sets is: lowercase letters, digits, and underscores only, starting with a letter, max 64 characters.

## Changes

**Single source of truth**
- `core/Constants.Name` — `PATTERN`, `MESSAGE`, `COMMENT_MAX_LENGTH`, `COMMENT_SIZE_MESSAGE`, `STORAGE_URI_PATTERN`, `STORAGE_URI_MESSAGE` as `const val`
- `server/util/NameValidator` — shared validator used by all controllers; throws `ResponseStatusException(400)` on mismatch
- `V3NameValidator` — now delegates to `NameValidator`

**V3 — create paths**
- `DatabaseController`, `TableController`, `AliasController`: name validation on POST only
- `TableCreateRequest`, `AliasCreateRequest`, `AliasUpdateRequest`: `@Pattern` and `@Size` annotations reference `Constants.Name`
- `TableCreateRequest.storage`: datastore URI regex updated to disallow uppercase
- Bug fix: `AliasUpdateRequest.table` was still using the old regex (allowed hyphens/uppercase)

**V2 — create paths only**
- `ServiceController`, `ServiceLabelController`, `ServiceAliasController`, `StorageController`: path variable validated on POST only
- `ServiceLabelController.copy`: new target name validated; existing source names are not

**HBase**
- `DatastoreHBaseService.createTable`: table name qualifier validated via `NameValidator`

**Tests**
- All hyphenated fixture names migrated to underscore style
- Explicit `hyphen returns 400` and `uppercase returns 400` test cases added per resource type
- `AliasControllerTest`: update-path validation tests for `table` field added
- `V3NameValidatorTest`: `my-db`, `MyDB`, `A` moved to invalid cases

## How to Test

Create with a hyphenated name → 400:
```sh
curl -X POST .../graph/v3/databases -d '{"database":"my-db","comment":""}' → 400
curl -X POST .../graph/v2/service/my-service -d '{...}' → 400
```

Read an existing legacy resource → 200:
```sh
curl .../graph/v3/databases/my-db → 200 (if it exists)
```

## ⚠️ Breaking Change — Deploy Prerequisites

This PR is a **breaking change for any code that creates metadata resources**.

Before deploying, all clients and pipelines that automatically create metadata (services, labels, aliases, databases, tables, storages) **must be audited and updated** to ensure they only submit names matching `^[a-z][a-z0-9_]{0,63}$`. Any pipeline submitting names with hyphens or uppercase will receive 400 after this lands.

Existing legacy-named resources remain readable, updatable, and deletable. Only creation is restricted — including new child resources under a legacy-named parent, which now return 400.

## Reviewer Notes

- Update/delete path variables are intentionally excluded from validation — existing legacy-named resources stay readable/updatable/deletable. Create paths do validate the parent name, so new children cannot be added under a legacy-named parent.
- `Constants.Name` is in the `core` module, which is already a dependency of both `engine` and `server`. No new module dependencies introduced.
- PR #399 (remove v2 Storage metadata) is based on this branch; `StorageController` changes here will be superseded by its deletion in #399.

## Checklist

- [x] Small, focused PR
- [x] Self-reviewed the diff
- [x] Tests updated and passing locally
- [x] Build and lint checks pass
- [x] Breaking change documented — clients/pipelines creating resources with hyphenated or uppercase names must be updated before deployment

## AI Assistance

  - Tool / model: Claude Sonnet 4.6

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
@@ -40,7 +40,6 @@ object Constants {
     object Name {
         const val PATTERN = "^[a-z][a-z0-9_]{0,63}$"
         const val MESSAGE = "name must start with a lowercase letter and contain only lowercase alphanumeric characters or underscores (max 64 chars)"
-        const val MAX_LENGTH = 64
 
         const val STORAGE_URI_PATTERN = "^datastore://[a-z][a-z0-9_]*/[a-z][a-z0-9_]*$"
         const val STORAGE_URI_MESSAGE = "storage must be in format datastore://<namespace>/<table> where each segment starts with a lowercase letter and contains only lowercase alphanumeric/underscore (e.g., datastore://my_namespace/my_table)"
--- a/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
@@ -41,6 +41,9 @@ object Constants {
         const val PATTERN = "^[a-z][a-z0-9_]{0,63}$"
         const val MESSAGE = "name must start with a lowercase letter and contain only lowercase alphanumeric characters or underscores (max 64 chars)"
 
+        const val COMMENT_MAX_LENGTH = 1000
+        const val COMMENT_SIZE_MESSAGE = "comment must be at most $COMMENT_MAX_LENGTH characters"
+
         const val STORAGE_URI_PATTERN = "^datastore://[a-z][a-z0-9_]*/[a-z][a-z0-9_]*$"
         const val STORAGE_URI_MESSAGE = "storage must be in format datastore://<namespace>/<table> where each segment starts with a lowercase letter and contains only lowercase alphanumeric/underscore (e.g., datastore://my_namespace/my_table)"
     }
--- a/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/Constants.kt
@@ -33,6 +33,16 @@ object Constants {
         const val BYTE_TRUE: Byte = 1
     }
 
+    /**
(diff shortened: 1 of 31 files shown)


em3s:
**Deploy prerequisites**

Before upgrading, confirm no existing resource violates the new policy — run via the Actionbase CLI:

```sh
actionbase -e "doctor names"
```

Expected: `No violations found.` Resolve every violation before the upgrade.

Two things `doctor names` won't catch — audit these separately:

1. **Legacy-named parent databases block new children.** Create paths validate the parent name, so even a policy-conforming table/alias cannot be created under a hyphenated/uppercase database. If such a database must keep receiving new tables, rename it — otherwise its children stay frozen.
2. **`/graph/v2/admin/hbase/...` `createTable` is not covered by this PR.** That admin endpoint writes directly to HBase without name validation. Any tooling creating tables through it must enforce the policy on its own.

Existing legacy-named resources remain readable, updatable, and deletable — only creation is restricted.

em3s:
Merging

PR #405, opened 2026-07-09
Log CDC publish failures with identifying context

eazyhozy:
## Summary

CDC publish failures were either logged with minimal context or silently dropped entirely, making it hard to tell which data failed to publish. This adds identifying details to the failure log so a CDC publish failure can be traced to a specific label/key/operation.

## Changes

- `V2BackedMessageBinding.writeCdc` (v3 mutation path): enrich the failure log with `label`, `key` (src:tgt, same as the Kafka partition key), `op`, `status`, `traceId`, and `requestId`, in addition to the exception.
- `Graph.mutate` (legacy path, two CDC write sites): these previously had no error callback at all — a CDC write failure was silently dropped with no log whatsoever. Added the same structured error logging.
- Added a test verifying `writeCdc` logs an ERROR with the above fields when the underlying `Cdc.write` fails.

## Call Flow

There are two independent call paths that both end up publishing CDC events, which is why this PR touches two files. Both paths converge at the same `Cdc.write(CdcContext)` interface, so the new logging helper is reused with the same format on both sides.

### 1. v3 data mutation path (regular edge/vertex traffic)

```
EdgeMutationController / VertexMutationController / MultiEdgeMutationController
  server/.../api/graph/v3/*MutationController.kt
  POST /graph/v3/{db}/{table}/edges, etc.
        |
        | mutationService.mutate(database, table, mutations, ...)
        v
MutationService.mutate(ctx, mutations, ...)
  engine/.../engine/service/MutationService.kt:60~75
    for each MutationEvent:
      engine.writeWal(ctx, event)              -- WAL first
      readModifyWrite(tb, key, sorted, ...)     -- actual HBase read-modify-write
        .doOnNext { result ->
          engine.writeCdc(ctx, sorted, result.status, ...)
        }
        |
        v
V2BackedEngine.writeCdc(ctx, events, status, before, after, acc)
  engine/.../v3/V2BackedEngine.kt:50
  = messaging.writeCdc(...) -- simple delegation
        |
        v
*** V2BackedMessageBinding.writeCdc(ctx, events, ...) ***      <- modified in this PR
  engine/.../v3/V2BackedMessageBinding.kt:49~75
    1) build CdcContext from events.last() (label, edge, op, status, before/after, requestId)
    2) cdc.write(cdcContext)
         .subscribeOn(Schedulers.boundedElastic())
         .subscribe(
            {},                                   -- success: no-op
            { logCdcWriteFailure(cdcContext, it) } -- failure: log (this PR)
         )
    NOTE: fire-and-forget -- fully decoupled from the mutation response.
    The mutation has already returned success by the time this runs,
    so a CDC failure is only ever visible through this log (-> AEM).
        |
        v
Cdc.write(CdcContext) = DefaultCdc.write(log)
  engine/.../cdc/DefaultCdc.kt:11
  = producer.produce(log)
        |
        v
Producer.produce(message)
  ProducerList -> KafkaProducer (engine wrapper)
  engine/.../producer/KafkaProducer.kt:49
  = client.send(Mono.just(message.toJsonBytes()))
        |
        v
KafkaClient.send(...) = SpringKafkaClient.send(...)  [production impl]
  server/.../client/kafka/SpringKafkaClient.kt:15~24
  = kafkaTemplate.send(SenderRecord.create(...)).then()
        |
        v
Kafka broker (topic configured for this deployment)
```

### 2. DDL (metadata schema) mutation path

```
ServiceController / ServiceLabelController / ServiceAliasController /
StorageController / AdminController
(v3: V3CompatService wraps the same DdlService instances as a v3-shaped API)
  server/.../api/graph/v2|v3/**Controller.kt
  POST/PUT/DELETE /graph/v2|v3/.../services|tables|aliases, etc.
        |
        | graph.serviceDdl / labelDdl / aliasDdl / storageDdl / queryDdl
        v                  (DdlService subclasses created in Graph.kt:154~162)
DdlService.create / update / delete
  engine/.../service/ddl/DdlService.kt:37~155
    checkCreate/Update/DeletePrecondition(...)
      .flatMap {
        val edge = request.toEdge(name)
        graph.mutate(label.name, label, listOf(edge), op, force = true, ...)
      }
        |
        v
*** Graph.mutate(alias, label, edges, operation, ...) ***      <- separate entry point from path 1
  engine/.../engine/Graph.kt:272
    wal.write(...)
      .then(
        label.mutate(edge, operation, alias, bulk, ...)
          .doOnNext { context ->
            cdc.write(context)
              .subscribeOn(Schedulers.boundedElastic())
              .subscribe({}, { logCdcWriteFailure(context, it) })       -- modified in this PR (was line 314)
          }
          .doOnError { cause ->
            val errorCdcContext = CdcContext(..., status = ERROR, message = cause.message)
            cdc.write(errorCdcContext)
              .subscribeOn(Schedulers.boundedElastic())
              .subscribe({}, { logCdcWriteFailure(errorCdcContext, it) }) -- modified in this PR (was line 332)
          }
      )
    NOTE: previously both subscribe() calls had no error callback at all --
    a CDC write failure here was silently dropped with no log whatsoever,
    which was even worse than the v3 path.
        |
        v
  (same as path 1 from here: Cdc.write -> DefaultCdc -> Producer -> SpringKafkaClient -> Kafka broker)
```

## How to Test

```
./gradlew :engine:test --tests '*V2BackedMessageBindingTest*'
```

Verify the new `writeCdc` test passes and asserts the log message contains `label`, `key`, `traceId`, `requestId`, and the original exception.

## AI Assistance

  - Tool / model: Claude Code (claude-opus-4-8)

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
@@ -311,7 +311,7 @@ class Graph(
                                         cdc
                                             .write(context)
                                             .subscribeOn(Schedulers.boundedElastic())
-                                            .subscribe()
+                                            .subscribe({}, { logCdcWriteFailure(context, it) })
                                     }.doOnError {
                                         val errorCdcContext =
                                             CdcContext(
@@ -329,7 +329,7 @@ class Graph(
                                         cdc
                                             .write(errorCdcContext)
                                             .subscribeOn(Schedulers.boundedElastic())
-                                            .subscribe()
+                                            .subscribe({}, { logCdcWriteFailure(errorCdcContext, it) })
                                     }.map { context ->
                                         MutationResultItem(
                                             status = context.status,
@@ -950,6 +950,23 @@ class Graph(
         internal val log = getLogger()
 
         val reactorLogger = reactor.util.Loggers.getLogger(Graph::class.java)
+
+        private fun logCdcWriteFailure(
+            cdcContext: CdcContext,
+            cause: Throwable,
+        ) {
+            log.error(
+                "CDC write failed. label={}, key={}:{}, op={}, status={}, traceId={}, requestId={}",
+                cdcContext.labelName,
+                cdcContext.edge.src,
+                cdcContext.edge.tgt,
+                cdcContext.op,
+                cdcContext.status,
+                cdcContext.edge.traceId,
+                cdcContext.requestId,
+                cause,
+            )
+        }
 
         @Suppress("LongMethod")
         fun create(
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedMessageBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/v3/V2BackedMessageBinding.kt
@@ -71,12 +71,29 @@ class V2BackedMessageBinding(
         cdc
             .write(cdcContext)
             .subscribeOn(Schedulers.boundedElastic())
-            .subscribe({}, { log.error("CDC write failed for {}/{}", ctx.database, ctx.table, it) })
+            .subscribe({}, { logCdcWriteFailure(cdcContext, it) })
     }
 
     companion object {
         private val log = LoggerFactory.getLogger(V2BackedMessageBinding::class.java)
         private const val DEFAULT_PRIMITIVE_VALUE = "0"
+
+        private fun logCdcWriteFailure(
+            cdcContext: CdcContext,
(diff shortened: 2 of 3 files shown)


eazyhozy:
Closing this PR based on offline review feedback from @em3s. The root cause of the original incident (silent CDC publish failure) is the `acks=1` Kafka producer setting — a leader election during publish can make the client believe the write succeeded when it didn't, so richer failure logging here cannot catch that case. Decision: fix this at the source by updating the Kafka producer config (`acks=all`, `min.insync.replicas=2`) in the deployment that hit this, rather than improving logging on the OSS side.

PR #406, opened 2026-07-10
Add GLOBAL scope to Topk aggregation

eazyhozy:
## Summary

Base: this PR stacks on #401 (`feature/per-entity-top-k-agg`), which introduces the topk metadata (`Group.aggregations.topk`) and the `POST /graph/v3/aggregations` endpoint.

Per-entity topk always derives the score row's entity from the edge's directed source, so different sources always map to different score rows. Global rankings (e.g. "most wished products across all users") need every matching edge to collapse into a single score-row prefix so one prefix scan returns the top-K. This PR generalizes the topk declaration so both cases (and property-backed ones) are expressed with the same two field references, and adds the refresh (re-aggregation) endpoints an external scheduler drives.

## Changes

**Topk declaration (`Group.aggregations.topk`)**

- `entity` / `rankedField` replace the earlier scope/rank-target enums: each names the field whose value scopes the ranking / is being ranked — `source`, `target`, or any property name. `entity` also accepts the `__global__` sentinel for a single ranking across all events. The defaults (`source` / `target`) preserve the per-entity behavior.
- `refreshAfterMillis` (renamed from `expireAfterMillis`) schedules re-aggregation. `table` declares only `score` — the refresh table is a fixed system table (`topk.refresh`, `AggregationConstants`), so `TopkTable.refresh` is removed.

**Score/refresh keys (`AggregationConstants`)**

- Score source key: `{database}.{table}|{topk}|{direction}|{entity}|{segment}`. The segment block always exists (`__all__` sentinel when no ranges) and stores the interpolated ranges raw for a GLOBAL topk.
- Blocks are delimited with `|`, which the ranges syntax (`:` `;` `,`) never uses, so raw segments stay readable. Events whose entity/rankedField value contains `|` are SKIPPED when refresh is enabled — the delimiter would mis-split the refresh-key parse.
- Refresh row: source = `xxhash(key) mod 2310` partition, target = `{scoreSource}|{rankedField}|{refreshAt}`, properties = `refreshAt` only. The key is the full coordinate; `parseRefreshTarget` restores it.

**Refresh endpoints (cron-pull, no CDC round-trip)**

- `GET /graph/v3/aggregations/refresh/entries?partition={n}&now={ts}&limit={l}` — scans one partition for due entries and returns parsed coordinates. Partition→worker mapping is the caller's job; the server stays stateless.
- `POST /graph/v3/aggregations/refresh` — re-aggregates each coordinate through the same internal path as the CDC endpoint, then deletes the processed refresh rows (idempotent; unresolvable entries are left for the next cycle to retry).

**Supersedes on the base branch** (agreed direction)

- The expire-row payload persistence (original endpoints / `directedSource` on rows) and non-bucket-group-field segmentation are replaced by the coordinate-key design: the row key alone re-derives everything re-aggregation needs, so rows carry no payload and no consumer parses refresh CDC.
- `GET /graph/v3/aggregations` no longer lists the system refresh table (the `refresh` flag is gone with it): with cron-pull refresh the streaming consumer only processes source-table CDC, so the entry had no remaining consumer.

**Fixes**

- Groups without any aggregation declaration no longer serialize an empty `"aggregations": {"topk": []}` block, keeping existing table-metadata responses unchanged.

## How to Test

- `./gradlew :core:test :engine:test :server:test`

## AI Assistance

  - Tool / model: Claude Code (Sonnet 5, Fable 5)


eazyhozy:
This PR is replaced with #401

PR #411, opened 2026-07-11
Remove dead system storage and V2 label constants

em3s:
## Summary

Remove two never-referenced system declarations. No behavior change.

- `localOnlyStorageEntity` / `localOnlyStorageName` — seeded into the metastore on boot, but no label references it as its `storage`. Dead since only `local_backed_metastore` / `metastore` / the HBase datastore back actual labels.
- `sysOnlineMetadataLabelName` (`"online_metadata"`, v1) — only the v2 name (`online_metadata_v2`) is used; the v1 constant has no reader.

## Changes

- Remove both declarations from `Metadata`.
- Drop the boot-time seeding of the local-only storage in `Graph`.
- Drop the matching `defaultStorages` / `defaultMetadata` entries in the test fixture.

## How to Test

`./gradlew :engine:test` — boot and metadata seeding are unchanged aside from the removed dead entries.

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/Graph.kt
@@ -1017,7 +1017,6 @@ class Graph(
 
             val storageEntities =
                 listOfNotNull(
-                    Metadata.localOnlyStorageEntity,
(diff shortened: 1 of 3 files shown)


em3s:
Optimistic

Issue #398, opened 2026-07-08
Remove v2 Storage metadata — enforce `datastore://` URI


em3s:
## Background

Retire the v2 `Storage` entity and enforce `datastore://<namespace>/<resource>` as the only storage reference. Part of metadata modernization (#347). This is a **breaking change**.

## Task

- [ ] Remove `Storage` entity and all related CRUD (create, get, sync)
- [ ] On startup, validate that every label's storage field starts with `datastore://` — fail fast if not

## Done When

`Storage` entity no longer exists. Server refuses to start if any label references a non-`datastore://` storage.

## Notes

- All existing labels must be migrated to `datastore://` storage references **before** upgrading.
- Related: #347

PR #410, opened 2026-07-11
Address system storage by URI scheme, drop JDBC path (#398)

em3s:
## Summary

Address system storage by URI scheme instead of named `Storage` entities, and drop the unused JDBC-only label path. Follows #399 (which blocked Storage creation) and continues retiring the v2 `Storage` model (#398).

Part of #398

## Scheme

A label's storage is now a self-describing `datastore://` URI resolved directly in `getStorage`:

- `datastore://__sys__/metastore` — RDB-backed system metastore (local + global)
- `datastore://<namespace>/<table>` — HBase datastore

Bare-name storage references still resolve via the `storages` map, so existing labels are unaffected.

## Changes

- `getStorage` collapses to a single scheme; `__sys__/metastore` resolves to a local-backed metastore storage.
- `Metadata` system labels reference `datastore://__sys__/metastore`; the LOCAL/JDBC Storage entities and their boot-time seeding are removed.
- Drop the JDBC-only materialization path — nothing runs a label on a standalone JDBC store in production. Removes `JdbcStorage`/`JdbcOptions`, the `is JdbcStorage` branch, the `StorageType.JDBC` materialize branch, and `JdbcHashLabel`'s create factory. `JdbcHashLabel` stays as the delegate base of `LocalBackedJdbcHashLabel`.

## How to Test

`./gradlew :engine:test` — system metadata boots and materializes over the metastore URI; existing labels with bare-name/HBase storage continue to resolve.

## AI Assistance

- [x] This PR was written largely with AI assistance.
  - Tool / model: Claude Opus 4.8

PR #412, opened 2026-07-11
Back metastore system labels by local-backed store, drop JDBC path

em3s:
## Summary

`query` and `alias` were the only system labels backed by the standalone JDBC metastore storage. Move them to the local-backed store and remove the now-dead JDBC path. Part of retiring the v2 `Storage` model (#398). The prepared-query feature is untouched — only its storage backend changes.

## Why this is behavior-preserving

The local-backed store in `useGlobal` mode is equivalent to the standalone JDBC store:

- **writes** go to the global store only (`useLocalStore = false`)
- **reads** merge local + global, but the local store stays empty, so the result equals global-only

So pointing `query`/`alias` at `local_backed_metastore` keeps their runtime behavior identical.

## Changes

- `query` / `alias` system labels now use `local_backed_metastore` instead of the `metastore` (JDBC) storage.
- Remove the `metastore` JDBC storage entity and its boot-time seeding.
- Drop the dead JDBC materialization path: `JdbcStorage` / `JdbcOptions`, the `StorageType.JDBC` materialize branch, the `is JdbcStorage` branch in `LabelEntity`, and `JdbcHashLabel`'s create factory. `JdbcHashLabel` stays as the delegate base of `LocalBackedJdbcHashLabel`.

## How to Test

`./gradlew :engine:test` — system metadata (including query/alias) boots and operates over the local-backed metastore; no label resolves to a JDBC storage.

## AI Assistance


em3s:
Superseded by #410, which reaches the same end state (JDBC path removed, query/alias consolidated) via datastore:// scheme URIs instead of keeping the named local_backed_metastore entity — the Storage-free direction for #398.

PR #413, opened 2026-07-13
Back the local metastore with an in-memory ByteArrayStore

em3s:
## Summary

Replace the H2-backed local store inside `LocalBackedJdbcHashLabel` with a single in-memory `ByteArrayStore`, removing H2 from the production runtime.

The store is created once in `Graph` and shared through `GraphDefaults.localStore`; the local side becomes a `ByteArrayIndexedLabel` while the global side stays a `JdbcHashLabel` on MySQL. `createDatabase` no longer provisions a separate local H2 database.

Builds on the `ByteArrayStore`-backed labels now on `main` (#421). Previously the tail of a 3-PR stack; #418 (dead inspector/endpoint removal) and #421 (the `ByteArrayStore` labels) have since merged, so this now targets `main` directly.

## Changes

- `LocalBackedJdbcHashLabel` now takes injected labels — local is a `ByteArrayIndexedLabel` over the shared `graph.localStore`, global stays `JdbcHashLabel`. `indexName=null` scans (`DdlService.getAll`) route to a synthetic `__default__` prefix-scan index.
- Drop `localMetastore` from `Graph` / `GraphDefaults`, add `localStore`, and remove the H2 branch of `createDatabase`.
- Keep H2 as a `runtimeOnly` dependency: the global metastore still falls back to in-memory H2 when no `metastoreUrl` is configured, so the server must keep the driver on the runtime classpath.
- `InitialSpec` verifies the system metadata by decoding hash edges straight from the shared local store; `LocalBackedJdbcHashLabelSpec` covers the overlay round-trip and the `indexName=null` scan routing; `GraphFixtures` is updated to build labels over the shared local store.

## How to Test

- `./gradlew build`
- `LocalBackedJdbcHashLabelSpec` covers the local write/read round-trip and `__default__` scan routing.
- Boots cleanly with the default profile (H2 metastore fallback) — verified `Started ServerApplicationKt` with `bootRun`.

## AI Assistance


em3s:
`LocalBackedJdbcHashLabel` was backed by two JDBC databases. This PR demonstrates that its local side can be replaced with a different storage -  one that is `IndexedLabel` (EdgeTable in v3), where it used to be a HashLabel - without any storage-backend infra, since the replacement is memory-backed. It's a good starting point for consolidating the metastore onto the `datastore://`.

em3s:
Merging

PR #415, opened 2026-07-14
Gate INSERT activating-merge behind the INSERT_MERGE feature

em3s:
## Summary

For a database enabled for `INSERT_MERGE` via `actionbase.feature-flags` (#432), INSERT becomes an activating merge: an omitted field keeps its existing value, an explicit null clears it, and only INSERT activates the row. This lets multi-source fan-in writers each INSERT the subset of fields they own without clobbering each other.

Without the feature (default), INSERT keeps snapshot semantics — ungated databases are unaffected, pre-gate WAL entries replay identically, and the completeness check keeps its original `IllegalArgumentException` (surfaced as `ERROR`).

```yaml
actionbase:
  feature-flags:
    - feature: INSERT_MERGE
      scope:
        databases: [fanin]
```

Behavior on an existing row, `email = "a@x.com"` (nullable):

| client input | INSERT (default, snapshot) | INSERT (INSERT_MERGE) | UPDATE |
|---|---|---|---|
| `{email:"b"}` | `"b"` | `"b"` | `"b"` |
| `{email:null}` | `null` (clear) | `null` (clear) | `null` |
| `{}` (omitted) | `null` (cleared) | **`"a@x.com"` (kept)** | kept |

## Changes

- `StateExtensions.transit(insertMerge)`: INSERT branches between merge and snapshot property handling. The active-row completeness check throws `InvalidMutationException` only under `insertMerge`; the default path keeps its original `IllegalArgumentException`, so ungated behavior is unchanged
- `MutationRequestValidator`: with the feature, INSERT relaxes to present-only nullability validation; explicit null on a non-nullable field still fails pre-WAL. Completeness of an activated row is enforced at transit (permanent)
- `MutationService`: resolves the feature once per request via the injected `FeatureFlags` (`has(Scope(database), INSERT_MERGE)`); a permanently invalid mutation surfaces as a per-item `INVALID` status (vs transient `ERROR`), detected across the exception cause chain
- `GraphConfiguration`: wires `MutationService` to `Graph.featureFlags`
- Tests: snapshot goldens unchanged from main; merge goldens in new `InsertMergeTest`; new `EdgeMutationInsertMergeTest` pins the gated pipeline end to end

## How to Test

- `./gradlew :core:test --tests "*InsertMergeTest"` — merge goldens
- `./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.MutationServiceSpec"` — per-item `INVALID` and fan-in merge through the RMW pipeline
- `./gradlew :server:test --tests "com.kakao.actionbase.server.api.graph.v3.EdgeMutationInsertMergeTest"` — E2E with the feature enabled via config

## AI Assistance

  - Tool / model: Claude Code (Opus 4.8, 1M context)

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/MutationEvent.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/MutationEvent.kt
@@ -20,5 +20,8 @@ sealed interface MutationEvent {
 }
 
 interface UnresolvedEvent {
-    fun createEvent(schema: ModelSchema): MutationEvent
+    fun createEvent(
+        schema: ModelSchema,
+        insertMerge: Boolean = false,
+    ): MutationEvent
 }
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/EdgeBulkMutationRequest.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/EdgeBulkMutationRequest.kt
@@ -14,11 +14,14 @@ data class EdgeBulkMutationRequest(
         val type: EventType,
         val edge: Edge,
     ) : UnresolvedEvent {
-        override fun createEvent(schema: ModelSchema): EdgeEvent {
+        override fun createEvent(
+            schema: ModelSchema,
+            insertMerge: Boolean,
+        ): EdgeEvent {
             require(schema is ModelSchema.Edge) { "Expected ModelSchema.Edge, but got ${schema::class.simpleName}" }
             val source = schema.source.type.cast(edge.source)
             val target = schema.target.type.cast(edge.target)
-            checkNonNullableFields(type, schema.properties, edge.properties)
+            checkNonNullableFields(type, schema.properties, edge.properties, insertMerge)
             val event =
                 Event.create(
                     type = type,
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/MultiEdgeBulkMutationRequest.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/MultiEdgeBulkMutationRequest.kt
@@ -14,10 +14,13 @@ data class MultiEdgeBulkMutationRequest(
         val type: EventType,
         val edge: MultiEdge,
     ) : UnresolvedEvent {
-        override fun createEvent(schema: ModelSchema): MultiEdgeEvent {
+        override fun createEvent(
+            schema: ModelSchema,
+            insertMerge: Boolean,
+        ): MultiEdgeEvent {
             require(schema is ModelSchema.MultiEdge) { "Expected ModelSchema.MultiEdge, but got ${schema::class.simpleName}" }
             val id = schema.id.type.cast(edge.id)
-            checkNonNullableFields(type, schema.properties, edge.properties)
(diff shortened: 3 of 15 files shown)


em3s:
### Follow-up concern: loss of payload-deterministic (pre-WAL) INSERT validation

Folding merge into `INSERT` surfaces a trade-off worth flagging — and it is likely the very reason v3 chose a snapshot `INSERT` in the first place.

Snapshot `INSERT` made completeness **payload-deterministic**: since every field is defined from the payload, an incomplete row could be rejected from the payload alone — before touching state, before the WAL. That bought:
- fail-fast pre-WAL rejection (a bad request never reaches the WAL/CDC stream),
- bulk all-or-nothing atomicity,
- the invariant that the stream never carries an `INSERT` that produces an incomplete active row.

Merge `INSERT` makes completeness **state-dependent** (an omitted field may be supplied by a prior write), so it cannot be validated from the payload — only in `transit`, after the WAL, per item. This cannot be recovered without a pre-WAL state read, which would break the WAL-first durability model. (Explicit-null and type errors are still caught pre-WAL; only the *completeness* dimension moves downstream.)

The only way to keep both properties is to keep them as two operations: snapshot `INSERT` (payload-deterministic, atomic validation) plus a separate activating-merge op (fan-in). Folding them into one sacrifices the atomic guarantee.

**Open question for reviewers:** does any WAL/CDC consumer or bulk client actually rely on "`INSERT` atomically produces a complete row before the WAL"? If yes, the two-operation design should be reconsidered. If no, this trade-off is acceptable and this PR stands as is.

em3s:
### INSERT / UPDATE / DELETE: v2 vs v3 (before) vs v3 (this PR), summarized

| | v2 | v3 (before) | v3 (this PR) |
|---|---|---|---|
| INSERT — value present | update | update | update |
| INSERT — explicit null | clear | clear | clear |
| INSERT — omitted | **kept** (merge) | **cleared** (snapshot) | **kept** (merge) ← now matches v2 |
| INSERT — activation | yes | yes | yes |
| UPDATE — value present | update | update | update (unchanged) |
| UPDATE — explicit null | clear | clear | clear (unchanged) |
| UPDATE — omitted | kept | kept | kept (unchanged) |
| UPDATE — activation | no | no | no |
| DELETE — field handling | all fields → `null` | all fields → `DELETED` sentinel | unchanged |
| DELETE — UNSET vs DELETED | not distinguished (both `null`) | distinguished | unchanged |
| DELETE — reorder guard | gated by `insertTs` | none | unchanged |
| DELETE — activation (deactivates row) | yes | yes | yes |

This PR touches exactly one cell — INSERT's "omitted" row, snapshot → merge — which is what brings INSERT back in line with v2. UPDATE and DELETE were already identical between v2 and v3 and stay that way. The two remaining structural v2/v3 differences are both on DELETE and are pre-existing, out of scope for this PR: v3 keeps `DELETED` distinct from `UNSET` (v2 doesn't), and v2 has a reorder guard on field clearing that v3 doesn't.

em3s:
Rollout note: #430 added the database-level feature gate, and this PR puts the INSERT merge semantics behind it — enabled per database via `actionbase.database-level-features`. Ungated databases keep the existing snapshot behavior byte-for-byte, so we can validate existing workloads and migrate incrementally instead of flipping the semantics everywhere at once.

em3s:
Optimistic

Issue #422, opened 2026-07-16
Unify store-backend acquisition on the StorageBackend URI factory


em3s:
## Background

Store-backend acquisition is inconsistent. On the live label path, backends are obtained ad-hoc: `Graph.create()` directly constructs `ByteArrayStore()` (twice — the local seed store and the consolidated overlay store), and the v2 labels take backend objects of different concrete types (`ByteArrayStore` vs `Mono<HBaseTables>` from `DefaultHBaseCluster`).

Meanwhile a URI-addressed factory already exists but is wired into nothing: `StorageBackend.getStorageTable("datastore://ns/table")` → `DatastoreUri.parse` → a `StorageTable` implementation (memory / mock-HBase / real HBase), selected by one config key (`hbase.type`). A third abstraction, `Datastore`/`DatastoreBucket`, is a dead duplicate (its only consumer reads a class name as a string; its memory bucket wraps the same `ByteArrayStore`).

Goal: converge store-backend acquisition on the existing `StorageBackend`/`StorageTable` seam so that a backend is requested by `datastore://` URI, and the metastore overlay's persistence (in-memory vs HBase) becomes a configuration choice rather than a code path.

Part of #347.

## Canonical target

- **Converge on `StorageBackend` / `StorageTable`** — the already-built URI-addressed factory with memory / mock-HBase / real-HBase implementations, URI parsing + validation, and `increment`.
- **Delete the `Datastore` / `DatastoreBucket` family** — dead duplicate.
- **Retire `DefaultHBaseCluster` into a `StorageBackend` implementation** — later, optional for this effort.

## Scope

In scope: the ByteArray / metastore label family (`ByteArrayHashLabel`, `ByteArrayIndexedLabel`, `LocalBackedJdbcHashLabel`). Its backend calls map 1:1 onto `StorageTable` (get / setnx / cad / batch / prefixScan / increment).

Out of scope: the HBase edge / indexed / cache label family. It needs capabilities `StorageTable` deliberately omits — a separate lock table, wide-row `ColumnRangeFilter` / qualifier-range Gets, and raw-HBase deferred batches. It stays on `HBaseTable` (sourced from the same backend). Forcing a full merge here has no #347 benefit and risks the hot read path.

## Plan (one PR per step; each independently shippable and reversible)

Critical path (#347 payoff):

- [x] **Step 0** — Delete the dead `Datastore` / `DatastoreBucket` / `MemoryDatastore` family. PR #423.
[...]

PR #425, opened 2026-07-16
Lock the exclusive-start scan pagination contract (pre-Step 2 of #422)

em3s:
## Summary

Preparatory step for #422 (Step 2): lock the scan pagination contract with a test before the ByteArray labels are reworked onto the `StorageTable` seam.

Both backend twins (`ByteArrayHashLabel`, `HBaseHashLabel`) already share the same scan contract — the start bound is **exclusive**: offset pagination resumes strictly after the cursor, so the boundary row is neither duplicated nor skipped. `StorageTable.scan`, on the other hand, uses the opposite HBase-native convention (start inclusive, stop exclusive). Pinning this now guards the upcoming reactive rewrite against a silent boundary shift (risk R1 in #422).

Part of #422.

## Changes

- Add `indexed - offset pagination resumes strictly after the cursor` to `AbstractLabelCompatibilityTest`, so it runs against both the `ByteArrayStore` and HBase backends. Test-only; no production change.

## How to Test

`./gradlew :engine:test --tests "*LabelCompatibilityTest"` — green on both backends; spotless clean.

## AI Assistance

- [x] This PR was written with AI assistance.

### DIFF
--- a/engine/src/test/kotlin/com/kakao/actionbase/v2/engine/label/AbstractLabelCompatibilityTest.kt
+++ b/engine/src/test/kotlin/com/kakao/actionbase/v2/engine/label/AbstractLabelCompatibilityTest.kt
@@ -147,4 +147,47 @@ abstract class AbstractLabelCompatibilityTest {
             .assertNext { assertEquals(listOf("v1", "v2"), it) }
             .verifyComplete()
     }
+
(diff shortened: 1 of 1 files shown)


em3s:
Optimistic

PR #430, opened 2026-07-19
Add database-level feature gate via application.yaml (INSERT_MERGE)

### LINKED ISSUE #429: Database-level feature management via application.yaml
## Background

To support fan-in writes (multiple independent sources each writing their own subset of fields onto the same row), INSERT is being switched to merge semantics (#415). We need a way to control the scope of this change instead of flipping the behavior everywhere at once.

Alternatives considered:

| Approach | Why rejected |
|---|---|
| Global flip (no gate) | Changes replay semantics of pre-deploy WAL entries; silently breaks any client relying on snapshot INSERT |
| Per-request flag | Write semantics become caller-chosen — contradicts the uncoordinated-writers premise; requires stamping the bit into WAL events, i.e. a wire change (the UPSERT we already rejected in #415) |
| Table metadata `features` field + dedicated API | The proper long-term path, but needs a metastore column, overlay support, API, and rollback compatibility — overkill for one feature on a few databases |
| Table-level application.yaml | Tables are routinely recreated with suffixed names and swapped behind aliases, so table-name-keyed config goes stale on every rotation; prefix matching could cope, but at that point database scope is simpler |
| Reusing comment/desc | Semantic bit in a free-text, freely editable field — no validation |
| **Database-level application.yaml (chosen)** | Minimal change, rolled out by deployment, audited via git |

Chosen shape:

```yaml
actionbase:
  database-level-features:
    - database: fanin
      features: [INSERT_MERGE]
```

This is a pragmatic minimal-implementation decision, not the optimal design — chosen to move fast. Once table metadata gains a `features` field (which survives renames and alias swaps), that becomes the canonical path and this gate is absorbed into it and removed.

## Task

Gate infrastructure only. Consuming the feature (the INSERT snapshot/merge branch in `StateExtensions.transit`, the pre-WAL validation branch, golden test parameterization) is out of scope here and lands with the #415 rework.

- [x] Add `TableFeature` enum (`INSERT_MERGE` — value only, no consumer)
- [x] server: `@ConfigurationProperties` binding — entry list (`database`/`features`), enum binding (typo fails boot), duplicate-database validation (fails boot), database name validation (fails boot), log effective entries at startup
- [x] Binding/validation tests

## Done When

- [x] `featuresOf(database)` resolves the configured feature set; unlisted databases resolve to an empty set
- [x] Invalid config fails at boot: unknown feature name, duplicate database entry, database name violating `^[a-z][a-z0-9_]{0,63}$`
- [x] Deploying with no entries changes nothing

## Notes

- #415 — will be reworked to consume this gate (INSERT merge semantics per database)
- See #415 comments for the v2/v3 INSERT semantics analysis

em3s:
## Summary

Adds a database-level feature gate configured via application.yaml. Infrastructure only — `INSERT_MERGE` is reserved but not consumed yet (that lands with the #415 rework).

```yaml
actionbase:
  database-level-features:
    - database: fanin
      features: [INSERT_MERGE]
```

Closes #429

## Changes

- `TableFeature` enum with `INSERT_MERGE`
- `ServerProperties`: bind `actionbase.database-level-features`, expose `featuresOf(database)`
- Boot fails on: unknown feature name, duplicate database, invalid database name

## How to Test

`./gradlew :server:test --tests "com.kakao.actionbase.server.configuration.ServerPropertiesTest"`

## AI Assistance

  - Tool / model: Claude Code (fable 5)


em3s:
Optimistic

PR #431, opened 2026-07-19
Disable flaky PerCellTTLTest

eazyhozy:
## Summary

`PerCellTTLTest` intermittently fails on the "Delete then Put" case: `OperationHelper.perform()` builds Delete/Put mutations without an explicit timestamp, so a Delete immediately followed by a Put can be assigned the same millisecond timestamp by HBase. Since a row tombstone covers cells with `timestamp <= tombstone timestamp`, the new Put gets shadowed — more likely to trigger on fast local machines than on CI.

Per discussion below, this test is really re-verifying HBase's own tombstone/TTL ordering rather than actionbase-specific logic, and that's already well-covered at the HBase level. Disabling it instead of chasing the timing race further.

## Changes

- `@Disabled` on `PerCellTTLTest`, with a pointer back to this PR for context.

## AI Assistance

  - Tool / model: Claude Code (Claude Fable 5)

### DIFF
--- a/engine/src/test/kotlin/com/kakao/actionbase/engine/experiments/hbase/PerCellTTLTest.kt
+++ b/engine/src/test/kotlin/com/kakao/actionbase/engine/experiments/hbase/PerCellTTLTest.kt
@@ -13,8 +13,10 @@ import kotlin.test.assertNull
 import org.apache.hadoop.hbase.client.Get
 import org.apache.hadoop.hbase.client.Table
 import org.apache.hadoop.hbase.util.Bytes
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.extension.ExtendWith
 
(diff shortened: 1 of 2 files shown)


em3s:
I think we should just disable it, assuming this is already well-tested at the HBase level. @eazyhozy

eazyhozy:
Agreed.

If we're dropping it, I'd lean toward **removing `PerCellTTLTest.kt` (and the `OperationHelper` fixture, its only user) entirely** rather than `@Disabled` — a disabled test with no re-enable plan is just dead code. Open to `@Disabled` if there's a reason to revisit it later.

@em3s which do you prefer?

em3s:
@Disabled

eazyhozy:
Disabled in 925fdf4.

em3s:
LGTM 👍

Issue #428, opened 2026-07-16
Introduce queue/v1: message queue over immutable edge tables


em3s:
## Summary

queue/v1 adds a message queue (also a **refresh / delay queue**) as an **engine capability** over the **immutable edge table** (#433/#434): the runtime (enqueue / poll) lives in `engine` (`com.kakao.actionbase.engine.queue`), with a thin `server/` HTTP layer and queue DDL over the v3 table API. A queue *is* an immutable edge table: `source` = partition, `target` = `id` (server-assigned ULID), a LONG `seq` property indexed **ASC** for per-partition order, and an opaque `value` (any JSON). Enqueue is a lock-free append; poll is a per-partition range scan over `seq`.

Message = `{ key, seq, value }`. The runtime lowers to the engine edge scan / mutation, admin to the v3 table DDL. Full design + request/response mappings live in the #437 design comment.

## Why immutable edge

Append-only, index-only, State-less — the storage shape of a message queue. Covers **enqueue (append)** and **poll (range scan + forward cursor)** directly.

## Model facts

1. **No Actionbase-level retention yet.** Only the storage backend's TTL applies today; Actionbase-level retention is a **TBD follow-up** (scan-and-delete).
2. **Delete is explicit.** Queue trim needs a **scan-and-delete** sweep — a **TBD follow-up** (not in #433).
3. **No cursor rewind.** The `offset` cursor only advances forward.
4. **Tombstones** from explicit deletes accumulate until compaction → **prefer bounded range scans**.
5. **gets / total out of scope.** Poll uses scan only.

## Concept mapping

| Queue | Immutable edge table |
|---|---|
| namespace | database |
| queue | table (`ImmutableEdge`) |
| partition = `floorMod(xxHash32(key), partitions)` | `source` (LONG) |
| message `id` (ULID) | `target` (STRING), unique per partition |
| `seq` (order / refresh due time) | `seq` LONG property + `Index([seq ASC])` |
| `value` (any JSON) | `value` STRING property (JSON blob) |
| enqueue | `MutationService.mutate` INSERT (append, no lock) |
| poll(partition, offset, until) | `scan(seq, start=partition)` with `seq:gt:offset` (log) / `seq:bt:offset+1,until` (refresh) |
| eviction | scan-and-delete (follow-up) |

## API

| Method | Path |
|---|---|
| POST | `/queue/v1/namespaces/{ns}/queues` |
| GET | `/queue/v1/namespaces/{ns}/queues/{queue}` |
[...]

PR #434, opened 2026-07-20
Add immutable edge table (index-only, append + scan)

### LINKED ISSUE #433: Add immutable edge table (index-only, scan / scan-delete)
## Summary

New table kind — **immutable edge** — persisting **only `EdgeIndex` rows, no edge `State`**. Append-oriented, read-via-scan. Maps to the existing `LabelType.IMMUTABLE_INDEXED`.

- **No point get** — `gets` rejected; `scan` is the only read.
- **First cut = Index + Group only** (Group backs count/sum-based topK). No Count, no Cache.
- **Delete is deferred** — scan-and-delete is descoped to a follow-up; the first cut is append + scan.

## Background

Needed for queue/v1 (#428). That design currently maps a queue onto a unique-Edge table, which carries an edge `State` row it never reads back — an index-only immutable table is the right storage primitive for a Kafka-style append log, and queue/v1 will build on it.

## Scope

Only v3 needs support — no new v2 mutation logic. The v3 binding still bootstraps from a v2 `HBaseIndexedLabel`, so `IMMUTABLE_INDEXED` materializes through the existing `INDEXED` path; all immutable behavior lives in v3, keyed off `ModelSchema.ImmutableEdge`.

## Tasks

- [x] Thread `ImmutableEdge` through the sealed hierarchies (schema, descriptors, strategy/builder, binding dispatch, create/update API).
- [x] Mutation emits index (+group) records only — no state / count / cache.
- [x] Reject `gets`; keep `scan` and `agg` working; reject the `total` scan feature.
- [x] Materialize `IMMUTABLE_INDEXED` via the `INDEXED` label path.
- [x] Tests: builder output + schema round-trip (unit); append → scan / gets rejected / count suppressed (HBase mini-cluster E2E).
- [ ] Scan-and-delete — deferred to a follow-up issue (out of scope here).

## Resolved decisions

- **Append index-key uniqueness**: no key change. Uniqueness = (indexValues + directedTarget); an identical key is an idempotent overwrite. With no State, immutable is pure append — no in-place update, no old-row cleanup.
- **Encoding record type**: reuse `EDGE_INDEX` / `EDGE_GROUP`. A distinct code would force a cross-lineage codec change (core + core-java + codec-java) and break the v3-only, scan-for-free property; a table is already namespaced by its label id.

## Follow-ups

- Scan-and-delete (delete during scan): the delete path for immutable tables.
- Count/Cache support if a use case needs them.

## Complexity

**High** — new kind across all parallel hierarchies, plus append-without-State.

em3s:
## Summary

Add a new **immutable edge** table kind — index-only, no edge `State`. Append-oriented, read-via-scan. Maps to the existing `LabelType.IMMUTABLE_INDEXED`. First cut supports Index + Group only (no Count, no Cache); scan-and-delete is deferred to a follow-up. Needed as the storage primitive for queue/v1 (#428).

Builds on the `KeyBased` strategy refactor (#435) and pairs with the bulk-encoder fix (#436).

Closes #433

## Changes

- Thread `ImmutableEdge` through the v3 sealed hierarchies (`ModelSchema`, `TableDescriptor`, `V3TableDescriptor`, `EdgeMutationStrategy` as a `KeyBased` object), placed right after `Edge`.
- Mutation emits index (+group) records only: skip the state `Put`, short-circuit `read` to initial (append-only), reject `gets` and the `total` scan feature.
- Append-only enforcement: non-INSERT mutations (UPDATE/DELETE) fail fast with a clear error instead of being silently ignored.
- Materialize `IMMUTABLE_INDEXED` through the existing `INDEXED` HBase label path; all immutable behavior lives in v3.
- Accept `ImmutableEdge` in the append event path and the create/update API + `V3MetadataConverter` round-trip.

## How to Test

- `./gradlew build` — full build incl. unit + HBase mini-cluster tests.
- `./gradlew :engine:test --tests "*ImmutableEdgeSpec"` — append → scan reads edges from index rows only, point get / non-INSERT rejected, count suppressed (HBase mini-cluster).
- `./gradlew :server:test --tests "*ImmutableEdgeE2ETest"` — full HTTP surface: create → append → scan; point get and non-INSERT return 400 (in-memory datastore backend).
- `./gradlew :core:test --tests "*EdgeMutationBuilderTest" --tests "*EdgeSchemaSerializationTest"` — builder emits index+group only; schema round-trips under the `immutableEdge` discriminator.

## AI Assistance

  - Tool / model: Claude Code

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/mutation/EdgeMutationBuilder.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/mutation/EdgeMutationBuilder.kt
@@ -37,6 +37,14 @@ object EdgeMutationBuilder {
         caches: List<Cache>,
     ): EdgeMutationRecords = buildWith(EdgeMutationStrategy.Edge, before, after, directionType, indexes, groups, caches)
 
+    fun buildForImmutableEdge(
+        before: EdgeStateRecord,
+        after: EdgeStateRecord,
+        directionType: DirectionType,
+        indexes: List<Index>,
+        groups: List<Group>,
+    ): EdgeMutationRecords = buildWith(EdgeMutationStrategy.ImmutableEdge, before, after, directionType, indexes, groups, emptyList())
+
     fun buildForMultiEdge(
         before: EdgeStateRecord,
         after: EdgeStateRecord,
@@ -50,19 +58,6 @@ object EdgeMutationBuilder {
         before: EdgeStateRecord,
         after: EdgeStateRecord,
     ): EdgeMutationRecords = buildWith(EdgeMutationStrategy.Vertex, before, after, DirectionType.OUT, emptyList(), emptyList(), emptyList())
-
-    /**
-     * Immutable edge: emits index (+group) records only — no count (strategy suppresses it)
-     * and no cache (caches are always empty here). The state record is still carried in
-     * [EdgeMutationRecords.stateRecord] but the binding skips the state `Put` for this kind.
-     */
-    fun buildForImmutableEdge(
-        before: EdgeStateRecord,
-        after: EdgeStateRecord,
-        directionType: DirectionType,
-        indexes: List<Index>,
-        groups: List<Group>,
-    ): EdgeMutationRecords = buildWith(EdgeMutationStrategy.ImmutableEdge, before, after, directionType, indexes, groups, emptyList())
 
     private fun buildWith(
         strategy: EdgeMutationStrategy,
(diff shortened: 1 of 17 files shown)


em3s:
Known limitation for future reference (not a blocker):

Immutable edge tables lose group-count idempotency. Because `read()` returns `State.initial` unconditionally, there's no state-transition dedup, so a retried/duplicated `INSERT` increments COUNT groups every time. Index rows stay idempotent; only group counts drift. Low impact today (append-only tables rarely use COUNT groups), and every fix is a trade-off — leaving it documented here.

PR #438, opened 2026-07-21
Add scan-and-delete for immutable edge tables

em3s:
## Summary

Add a **scan-and-delete** primitive for immutable edge tables and expose it end-to-end — the building block queue/v1 uses to bound its log. Both a consumer **commit** (poll → process → commit, #439) and plain retention reduce to the same thing: delete a processed prefix by scanning an index range and deleting the matched rows.

An immutable edge table persists *only* index rows (no State row, no count), so deleting the scanned index keys is a complete delete. HBase has no native range delete, so this scans the matching index rows and issues per-row point deletes (bounded by `limit`; callers loop to trim a larger range).

Builds on the immutable-edge constraint from #440 (already merged to `main`): one index in one direction is exactly what makes "deleting the scanned index keys is a complete delete" true by construction.

## Changes

Engine primitive:
- `IndexedLabelMixin.scanAndDeleteIndexedEdges(scanFilter)` — scans one page of a range and deletes the matched index rows via the existing `deleteIndexedEdges`, returning the count deleted.
- `TableBinding.scanDelete(index, start, direction, limit, ranges)` — new binding op; `V2BackedTableBinding` implements it **gated to immutable edges** (`require(isImmutable)`), `NilTableBinding` is a no-op.
- `MutationService.scanDelete(...)` exposes it on the write path and **caps `limit` at `MAX_SCAN_DELETE_LIMIT` (1000)** so one call cannot scan-and-delete an unbounded range in a single RPC batch; callers loop to drain a larger range. Enforced at the engine, so every caller is bounded.

User-facing surface:
- `EdgeMutationController`: `DELETE /graph/v3/databases/{database}/tables/{table}/edges/scan/{index}` (`start`, `direction`, `limit`, `ranges`) → `EdgeScanDeleteResponse{database, table, index, deleted}`. `limit` is required (destructive op — force explicit intent). Blocked in read-only mode by the existing `ReadOnlyRequestFilter` (non-GET).

Deliberately immutable-only: for a mutable indexed edge, deleting only index rows would leave the State row and count dangling.

## How to Test

- `./gradlew :engine:test --tests "*ImmutableEdgeSpec"` — includes a case: append 3 rows, `scanDelete` `seq <= 1001`, assert 2 deleted and only the newer row remains.
- `./gradlew :server:test --tests "*ImmutableEdgeE2ETest"` — includes E2Es: append then `scanDelete` deletes the matched page and leaves the rest; an over-cap `limit` is rejected with 400.
- `./gradlew :engine:spotlessCheck :server:spotlessCheck`.

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/binding/TableBinding.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/binding/TableBinding.kt
@@ -57,6 +57,18 @@ interface TableBinding {
         features: List<String>,
     ): Mono<DataFrame>
 
+    /**
+     * Scans one page of `index`/`ranges` and deletes the matched rows, returning the count deleted.
+     * Immutable edge tables only — their index rows are the whole record. Bounded by `limit`.
+     */
+    fun scanDelete(
+        index: String,
+        start: Any,
+        direction: Direction,
+        limit: Int,
+        ranges: String?,
+    ): Mono<Int>
+
     fun seek(
         cache: String,
         start: List<Any>,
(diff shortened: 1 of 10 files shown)


em3s:
Optimistic

PR #439, opened 2026-07-21
Add commit (poll -> process -> commit) to queue/v1

em3s:
## Summary

Add **commit** to queue/v1: a consumer polls a partition, processes the batch, then commits up to that batch's offset — every message with `seq <= offset` is deleted. `DELETE /queue/v1/namespaces/{ns}/queues/{queue}/partitions/{partition}/messages?offset=X`.

This completes the **poll → process → commit** lifecycle and doubles as retention/eviction. An immutable edge keeps no per-message ack state, so there is a single operation: the delete *is* the committed position. `offset` mirrors poll's forward cursor, so `commit(offset = polledPage.offset)` deletes exactly the prefix just polled.

Stacked on #438 — base is `feat/immutable-edge-scan-delete`; review/merge that first. `commit` is the queue-level use of #438's scan-and-delete primitive. (The immutable-edge constraint from #440 is already in `main`.)

Single consumer per partition: committing deletes the prefix, so it assumes one logical consumer advancing a partition (Kafka-style), not independent consumers sharing one.

## Changes

- `QueueService.commit(namespace, queue, partition, offset)` — loops `MutationService.scanDelete` in pages of `COMMIT_BATCH` until the `seq <= offset` range is drained, returning the total deleted; validates the partition is in range.
- `MessageController`: `DELETE .../partitions/{partition}/messages?offset=X` → `QueueCommitResponse{namespace, queue, partition, committed}`.
- `ReadOnlyRequestFilter` classifies the endpoint as a write (blocked in read-only mode).

## How to Test

- `./gradlew :server:test --tests "*queue.v1.*"` — includes a lifecycle E2E: enqueue seq 1..5, poll a batch of 3, `commit(offset = page.offset)`, assert 3 deleted and a re-poll from the start returns only 4/5, then commit the rest and the partition drains.
- `./gradlew :engine:spotlessCheck :server:spotlessCheck`.

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/queue/QueuePayloads.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/queue/QueuePayloads.kt
@@ -42,3 +42,10 @@ data class QueuePartitionsResponse(
     val queue: String,
     val partitions: Int,
 )
+
+data class QueueCommitResponse(
+    val namespace: String,
+    val queue: String,
+    val partition: Int,
+    val committed: Int,
+)
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/queue/QueueService.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/queue/QueueService.kt
@@ -96,6 +96,34 @@ class QueueService(
         }
     }
 
+    fun commit(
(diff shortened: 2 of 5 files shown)


em3s:
Optimistic

PR #440, opened 2026-07-21
Constrain immutable edge tables to one index and one direction

em3s:
## Summary

Establish the **constraints for an immutable edge table**: exactly **one index** in **one direction** (OUT or IN, not BOTH).

Surfaced while implementing scan-and-delete. An immutable edge persists *only* index rows (no State row, no count), so the index rows **are** the record — and an append writes one row per `(index x direction)`. Eviction scans a **single** index in a **single** direction and deletes only those rows. Any row keyed by another index, or by the opposite direction, would be orphaned: unreachable to trim via the scanned path, yet still returned by scans that reach it.

Constraining an immutable edge to one index and one direction keeps "the index rows are the record" true by construction, so eviction stays a complete delete. This is a prerequisite for the immutable-edge scan-delete work.

## Changes

- `ModelSchema.ImmutableEdge` — `init` block enforces two invariants:
  - `indexes.size <= 1` (message names the offending indexes)
  - `direction != BOTH`

  Enforced at the model boundary, so it holds for every entry point (table create/update, deserialization, direct construction) and yields a 400 on the create/update path.

## Compatibility

The check runs on every construction, including reads (`toV3ModelSchemaImmutableEdge` reconstructs the schema from stored metadata). Immutable edge is a new table type and queue/v1 (single `seq_asc` index, OUT) is compliant, so no existing table is affected; a pre-existing table that violated the rule would now fail to read back.

## How to Test

- `./gradlew :core:test --tests "*EdgeSchemaSerializationTest"` — includes cases asserting a two-index and a BOTH-direction `ImmutableEdge` construction each throw.
- `./gradlew :server:test --tests "*ImmutableEdgeE2ETest"` — includes E2Es: creating an immutable table with two indexes, or with BOTH direction, returns 400.
- `./gradlew :core:spotlessCheck :server:spotlessCheck`.

## AI Assistance

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/ModelSchema.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/metadata/common/ModelSchema.kt
@@ -46,7 +46,17 @@ sealed class ModelSchema : AbstractSchema {
         val indexes: List<Index> = emptyList(),
         val groups: List<Group> = emptyList(),
     ) : ModelSchema(),
-        AbstractSchema by Schema(properties.associate { it.name to it.nullable })
+        AbstractSchema by Schema(properties.associate { it.name to it.nullable }) {
+        init {
+            // An append writes one index row per index, and those rows are the whole record (no State
+            // row). Eviction scans a single index and deletes only its rows, so a second index's rows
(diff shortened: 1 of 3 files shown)


em3s:
Merge

PR #441, opened 2026-07-21
Immutable edge tables do not emit CDC

em3s:
## Summary

Immutable edge tables no longer emit CDC. An immutable edge is an append-only log consumed by scan/poll (it backs queue/v1), not a CDC stream — the log itself *is* the change history, so streaming a parallel CDC event per append is redundant.

This also removes an asymmetry surfaced while adding scan-and-delete (#438): appends emitted CDC while evictions — which delete storage directly — did not. Now neither does, so an immutable table is consistently CDC-free.

WAL is unaffected: `writeWal` runs before the CDC step, so durability and async consumption are unchanged. Only the downstream CDC stream is skipped, and only for immutable edges.

## Changes

- `MutationService.mutate` — skip `engine.writeCdc(...)` when the table's schema is `ModelSchema.ImmutableEdge`. Mutable edge / multi-edge / vertex paths are untouched.

## How to Test

- `./gradlew :engine:test --tests "*ImmutableEdgeSpec"` — includes a case: append succeeds (CREATED) yet emits no CDC for the table.
- `./gradlew :engine:test --tests "*GraphSystemAsyncSpec" --tests "*CdcSpec"` — non-immutable mutations still emit CDC (regression guard).
- `./gradlew :engine:spotlessCheck`.

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/engine/service/MutationService.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/engine/service/MutationService.kt
@@ -4,6 +4,7 @@ import com.kakao.actionbase.core.edge.MutationEvent
 import com.kakao.actionbase.core.edge.MutationKey
 import com.kakao.actionbase.core.edge.UnresolvedEvent
 import com.kakao.actionbase.core.edge.payload.MutationResult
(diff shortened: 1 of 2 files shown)


em3s:
Optimistic

PR #443, opened 2026-07-22
Support serializable date buckets in codec-java

eazyhozy:
## Summary

At the time of this change, neither `main` nor `codec-java/0.4.x` defines date bucket metadata, and the published codec-java 0.4.0 artifact has no `Bucket` type or `Group.Field.bucket`. When metadata contains a date bucket, strict deserializers can reject the unknown `bucket` property and fail to load the metadata, while tolerant deserializers can silently discard the bucket definition. In the latter case, downstream bulk-load or distributed processing can group by raw timestamp values instead of the formatted bucket values expected by the metadata.

This PR adds date bucket metadata support to codec-java group fields and preserves compatibility with 0.4.0-shaped metadata where `bucket` is missing or explicitly null. It also keeps derived `ZoneId` and `DateTimeFormatter` state out of Java serialization and reconstructs that state after deserialization so nested `LabelDTO` metadata remains usable.

Closes: N/A

This PR is a prerequisite for #397 and should be merged first.

## Changes

- Add date bucket metadata and application to `Group.Field`.
- Keep optional buckets compatible with metadata where `bucket` is missing or null, including strict creator-null checks.
- Keep derived `ZoneId` and `DateTimeFormatter` state out of Java serialization and reconstruct it after deserialization.
- Cover value-unit conversion, missing/null/date bucket deserialization, and nested `LabelDTO` Java serialization round trips.

## How to Test

- [x] `./gradlew :codec-java:test --tests com.kakao.actionbase.v2.core.metadata.BucketTests --tests com.kakao.actionbase.v2.core.metadata.LabelDTOTests`
- [x] `./gradlew :codec-java:spotlessCheck`

## AI Assistance

  - Tool / model: Codex / GPT-5.6 SOL

---
🤖 Generated with Codex (gpt-5.6-sol)

### DIFF
--- a/codec-java/src/main/java/com/kakao/actionbase/v2/core/metadata/Bucket.java
+++ b/codec-java/src/main/java/com/kakao/actionbase/v2/core/metadata/Bucket.java
@@ -1,0 +1,152 @@
+package com.kakao.actionbase.v2.core.metadata;
+
+import com.kakao.actionbase.v2.core.types.DataType;
+
+import java.io.Serializable;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/**
+ * Ported from V3 {@code com.kakao.actionbase.core.metadata.common.Bucket}. Only the {@code date}
+ * bucket is supported, and only {@link #apply(Object)} (bulk-encoding time bucketing) is ported —
+ * {@code handleQueryValue} is query-time only and unused by bulk encoding.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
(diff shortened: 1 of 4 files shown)


eazyhozy:
This PR adds serializable date bucket metadata to codec-java so that it can be e2e-tested via internal bulkload pipeline. This should land before #397. 

Could you review this PR? @em3s @zipdoki

zipdoki:
@eazyhozy 
Could you run an actual bulkload with this branch to confirm it works end-to-end before merging?

eazyhozy:
@zipdoki Sure. I published codec-java from this branch and rebuilt the downstream Spark bulk-load jobs against it. I am replaying an actual bulk load through the metadata-creation stage. I will report whether the replay completes through that stage and whether the newly created table preserves the existing schema, especially its groups and date-bucket metadata (name, unit, timezone, and format). I will also run a historical AGG query; an empty result is expected at this point because EdgeGroup bulk-load encoding is introduced separately in #397.

eazyhozy:
@zipdoki Validation completed successfully:

- Replayed a downstream Spark bulk load through table creation using codec-java built from this branch.
- The newly created table schema matches the existing table schema exactly.
- All groups were preserved, and every date bucket retained its type, name, unit, timezone, and format metadata.
- Raw edge data was loaded and directly queryable.
- A historical AGG query returned an empty result as expected because EdgeGroup bulk-load encoding is introduced separately in #397.

This confirms that `LabelDTO` can be serialized and deserialized through the downstream bulk-load flow without losing group or date-bucket metadata.

zipdoki:
Optimistic Merge

PR #459, opened 2026-07-24
Back datastore HASH labels with an IndexedLabel on the V3 path

em3s:
## Summary

A HASH label on datastore storage can't be mutated on the V3 path. Every edge mutation fails with `This Label (<db>.<table>, class ...DatastoreHashLabel) is not indexed or not supported for edge mutation`.

`V2BackedEngine.getTableBinding` only accepts labels in the `HBaseIndexedLabel` family. HBase HASH labels clear that gate through `HBaseHashOnlyIndexedLabel` (#296), but the datastore branch of `LabelEntity.materialize` produced `DatastoreHashLabel`, which extends `HBaseHashLabel` directly and sits outside that family. The label materializes and passes DDL, then fails only at mutation time with a `400`. This reproduces identically on `v0.4.6` and `main` — the datastore HASH path was never wired through the V3 table binding.

The fix adds `DatastoreHashOnlyIndexedLabel`, the datastore mirror of `HBaseHashOnlyIndexedLabel`: an `HBaseIndexedLabel` with no indices whose `scan`/`cache` throw. With no indices the indexed mutation path writes only the base hash rows, so the on-disk format stays identical to a plain hash label and existing datastore HASH tables keep working with no data migration.

A backport to `0.4.x` follows in a separate PR.

## Changes

- Add `DatastoreHashOnlyIndexedLabel` and route `HASH + DatastoreStorage` to it in `LabelEntity.materialize`
- Remove the now-unused `DatastoreHashLabel`
- Add `DatastoreHashLabelSpec` covering materialization type, `getTableBinding`, edge mutation, and `get`

## How to Test

```
./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.DatastoreHashLabelSpec"
```

Full `:engine:test` passes locally.

## AI Assistance


em3s:
Optimistic

PR #461, opened 2026-07-24
Clarify getTableBinding error message to cover the query path

em3s:
## Summary

`V2BackedEngine.getTableBinding` serves both the query and the mutation path, but its `UnsupportedOperationException` said the label `is not indexed or not supported for edge mutation`. When the error fires on a GET request the message points the diagnosis at the wrong path. Reword it to name the V3 table binding and state that both query and edge mutation are affected.

## Changes

- Reword the `getTableBinding` error message in `V2BackedEngine` to cover both the query and the mutation path

## How to Test

```
./gradlew :engine:test --tests "com.kakao.actionbase.v2.engine.v3.*"
```

## AI Assistance

  - Tool / model: Claude Code (Fable 5)


em3s:
Merge

PR #469, opened 2026-07-24
Resolve an omitted datastore:// namespace to the backend default

### LINKED ISSUE #468: Define the datastore:// storage URI scheme
## Context

A table's storage ref is already a `datastore://` URI, but the scheme was never written down — this pins it as a contract.

The gap is the namespace. `actionbase.hbase.namespace` in `application.yaml` (`ab_local`, `ab_prod`) is the per-deployment namespace, but acquisition ignores it — it parses the namespace out of the URI and passes it straight to `TableName.valueOf`. So migrate bakes a concrete namespace into every ref (`HBASE → datastore://<namespace>/<table>`), tying persisted metadata to one environment. `DatastoreUri.parse` already allows an empty namespace, so the scheme looks meant to support omission — nothing fills the blank in.

Fix: adopt Hadoop's `fs.defaultFS` model — namespace optional, an omitted one resolves to the backend's configured default.

## Decision

**Format & naming**
- [x] `datastore://<namespace>/<table>`; single-segment (`datastore://<table>`, no `///`) rejected; table segment required (`DatastoreUri.parse`)
- [x] Segments start with a lowercase letter and contain only lowercase alphanumeric/underscore (`[a-z][a-z0-9_]*`) — the same rule every other name follows (`Constants.Name.PATTERN`)
- [x] The fully delimited `__x__` form is reserved for system namespaces. User input cannot express it, since the letter-start rule already excludes it; the internal parser accepts it so the engine can mint `datastore://__sys__/metastore`
- [x] `datastore://__sys__/metastore` reserved (`EngineConstants`)

**Namespace resolution (Hadoop `fs.defaultFS` model)**
- [x] Namespace may be omitted (`datastore:///<table>`), and the DDL boundary accepts the omitted form so a table created through the API can carry a portable ref
- [x] An omitted namespace resolves to the backend's `defaultNamespace`; `StorageBackend.getStorageTable(uri)` and `DefaultHBaseCluster` substitute it for an empty namespace
- [x] `HBaseStorageBackend` supplies its configured namespace as the default (previously discarded)
- [x] memory/mock backends default to `"default"`

**Persistence policy**
- [x] Metadata omits the namespace (`datastore:///<table>`) so refs stay portable; an explicit namespace is written only for a deliberate cross-namespace ref
- [x] The migrate rewrite emits the omitted form and reports the source namespaces it drops, warning when more than one appears

## Amendment (2026-07-31)

I rewrote the two naming bullets. They described the parser as it stood — any lowercase alphanumeric/underscore segment, system names reserved by a `__` prefix — rather than a rule. Every other name in the API starts with a lowercase letter, so the storage segment was the lone exception, and a `__` prefix admits half-delimited names like `__sys` that belong to neither side. Requiring `__x__` makes the reserved set a shape, and letter-start keeps it out of user input without a lookahead.

This gives up `datastore:///_expire`. Nothing mints that today and the name could never have come through DDL, so if we want that spelling later it needs a decision here first. I also made the DDL regex accept the omitted form — the persistence policy already called it the portable one, but the regex required a namespace.

#386

## Consequences

- Metadata becomes environment-portable — the same omitted-form ref resolves against `ab_local` locally and `ab_prod` in prod, no rewrite. Namespace is deployment config again, not baked into storage.
- Explicit (`datastore://ab_local/t`) and omitted (`datastore:///t`) are different refs — one absolute, one relative to the configured default. That distinction is the point of the scheme.

em3s:
## Summary

The namespace in a `datastore://` ref is a per-deployment config value (`actionbase.hbase.namespace` → `ab_local`, `ab_prod`), but acquisition passed the URI's namespace straight to `TableName.valueOf`. So the migrate CLI baked a concrete namespace into every rewritten ref and pinned the metadata to one environment, and the `defaultNamespace` the factory already captured was never read.

This implements the `datastore://` scheme decision from #468: adopt Hadoop's `fs.defaultFS` model, where an omitted namespace (`datastore:///<table>`) resolves to the backend's configured default at acquisition. An explicit namespace still resolves exactly as before, so existing refs are untouched.

I amended the naming rules in #468 while writing this. Segments start with a lowercase letter like every other name, and the reserved form is the fully delimited `__x__`.

Closes #468

## Changes

- `StorageBackend` carries a `defaultNamespace`; `getStorageTable(uri)` substitutes it for an empty parsed namespace.
- `HBaseStorageBackend` supplies the namespace it previously only `requireNotNull`d, and now rejects a blank one — blank would leave an omitted-namespace ref resolving to HBase's own `default` rather than the deployment's. Backends that serve one unnamed space share `StorageBackend.DEFAULT_NAMESPACE`.
- `DefaultHBaseCluster` fills an empty namespace with its configured namespace on the data-table resolution path. Its parsing now delegates to `DatastoreUri.parse` — the data-table path had a private copy of the split-and-require logic, so the scheme's rules now live in one parser.
- `STORAGE_URI_PATTERN` accepts the omitted form (`datastore:///<table>`) and holds each segment to the letter-start rule, which leaves the reserved `__x__` form out of reach of user input so it can't target `datastore://__sys__/metastore`.
- The migrate CLI rewrite emits the omitted form instead of baking the namespace. `migrate plan` reports the source namespaces it drops — counting only the storages a planned label references — and rejects a `conf.tableName` the scheme disallows, since legacy storage entities were never held to the name rule.
- Tests: omitted-namespace resolution on all three backends and on `DefaultHBaseCluster`, a blank namespace rejected at `create`, the pattern's accept/reject set (including the `__x__` reservation and the letter-start rule), and the migrate URI shape.

One behavior change rides along. `DefaultHBaseCluster.getTable(uri)` validated nothing before, so a ref with characters the scheme disallows reached `TableName.valueOf`; it now fails at resolution. Nothing creatable through DDL is affected, and the migrate rewrite rejects such a name in the plan phase rather than emitting it.

## How to Test

```
./gradlew :core:test :engine:test
cd cli && go test ./internal/command/
```

## AI Assistance

### DIFF
--- a/cli/internal/command/migrate.go
+++ b/cli/internal/command/migrate.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/kakao/actionbase/internal/client"
@@ -73,15 +74,20 @@ func (m *Migrate) plan(outputPath string) *model.Response {
 	// with the reason if a label actually references one.
 	storageURIByName := map[string]string{}
 	storageErrByName := map[string]error{}
+	droppedNamespaces := map[string]struct{}{}
 	if resp := m.client.GetStorages(); !resp.IsError() {
 		for _, s := range resp.Body.Content {
 			if uri, err := storageToDatastoreURI(s); err != nil {
 				storageErrByName[s.Name] = err
 			} else {
 				storageURIByName[s.Name] = uri
+				if ns := confString(s.Conf, "namespace"); ns != "" {
+					droppedNamespaces[ns] = struct{}{}
+				}
 			}
 		}
 	}
+	reportDroppedNamespaces(droppedNamespaces)
 
 	var plan []migrationEntry
 	add := func(path string, body map[string]any, label string) {
@@ -241,6 +247,27 @@ func storageToDatastoreURI(s clientModel.StorageEntity) (string, error) {
 		return "", fmt.Errorf("HBASE storage %q conf is missing tableName", s.Name)
 	}
 	return fmt.Sprintf("%s/%s", datastoreURIPrefix, tableName), nil
+}
+
+// reportDroppedNamespaces surfaces the source namespaces the rewrite drops.
+// The rewritten refs omit the namespace, so the server resolves them to the
+// target deployment's configured namespace. If more than one source namespace
+// appears, tables that were not in the target's default namespace would be
+// silently repointed — the operator must verify before applying.
+func reportDroppedNamespaces(namespaces map[string]struct{}) {
+	if len(namespaces) == 0 {
+		return
+	}
+	names := make([]string, 0, len(namespaces))
+	for ns := range namespaces {
+		names = append(names, ns)
+	}
+	sort.Strings(names)
+	util.Print("Note: rewritten refs omit the source namespace and resolve to the target deployment's configured namespace.\n")
+	util.Print("      Source namespaces seen: %s\n", strings.Join(names, ", "))
+	if len(names) > 1 {
+		util.Print("      WARNING: multiple source namespaces — tables not in the target's default namespace will be repointed. Verify before apply.\n")
+	}
 }
 
 func confString(conf map[string]any, key string) string {
--- a/cli/internal/command/migrate.go
+++ b/cli/internal/command/migrate.go
@@ -228,16 +228,19 @@ func tableBody(t clientModel.TableEntity, uriByName map[string]string, errByName
 // URI. The engine resolves every datastore:// URI except the reserved
 // __sys__ metastore as an HBase namespace/table, so non-HBASE storages have
 // no URI to mint — they return an error instead.
+//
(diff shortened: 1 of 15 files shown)


em3s:
Two notes for review.

The naming rules in #468 changed while I was finishing this. Segments now start with a lowercase letter like every other name in the API, and the reserved form is the fully delimited `__x__` rather than a `__` prefix. I amended the issue body before pushing so the decision record and the regex agree. Against `main` the pattern is still purely additive — it accepts the omitted form, and nothing that used to pass now fails.

One thing I left alone. `DefaultStorageBackendFactory` keeps its own `defaultNamespace` (`properties["namespace"] ?: "default"`), so two places now answer "what does an omitted namespace resolve to". Nothing in production reads the factory's copy, and with `type=memory` the two disagree — the factory reports the configured namespace while the backend reports `default`. I'd rather not churn it here: the planned follow-up that builds the storage provider from configuration drops the dependency on that global entirely and leaves the factory as test infrastructure, which is where the duplicate should go away.

em3s:
Optimistic

Issue #370, opened 2026-06-02
Introduce the Control Plane role: topology, orchestration, authorization, audit


em3s:
## Background

Actionbase was originally designed for a single-tenant, single-cluster deployment model.

Today production environments are more complex:

* Multiple tenants
* Active/Standby deployments
* Dedicated read clusters
* Environment-specific policies (dev/test/prod)
* User-level authorization
* Audit requirements
* Operational workflows such as cleanup

In these environments, operational actions such as DDL, cleanup, and cluster management must be coordinated across multiple Actionbase clusters.

Every operational client — Console, CLI, scripts — carries its own table of which hostname is which, and each one reimplements the same orchestration: the order of steps, the preconditions that make a step safe, the active/standby fanout, the queue that retries what failed. None of it is reusable, and it couples clients to deployment topology.

That logic belongs in one place: the Control Plane.

## Architecture

**Revised in #480: one binary, two roles — not a separate module.**

This issue originally called for a new `control-server` module. The dependency argument for splitting turned out to be about image size rather than correctness, so a config property decides what an instance is instead:

```
actionbase.role=DATA      Query, DML and DDL against this instance's own cluster. The default.
actionbase.role=CONTROL   Operational APIs spanning clusters: topology, orchestration, audit.
```

One artifact, one image, one release. The two HTTP surfaces are exclusive — a `DATA` instance refuses `/control`, a `CONTROL` instance refuses `/graph/v2`, `/graph/v3` and `/queue/v1` — so the runtime isolation is what two modules would have given, without a second release to maintain.

```text
Console / CLI
        ↓
  role=CONTROL      topology, authorization, sequencing, audit
        ↓
  role=DATA         query, DML, DDL, per-cluster invariants
        ↓
      HBase
```

An existing deployment that sets no role keeps behaving exactly as before.

## Control Plane vs Data Plane

The terms are used from an architectural perspective.

* **Data Plane** — data access, cluster-local request execution, and the invariants of a single cluster.
* **Control Plane** — topology, authorization, operational policy, orchestration across clusters, and auditing.

DDL is inherently a control-plane concern, but DDL *execution* stays in the data plane. The control plane decides whether, in what order, and against which clusters.

## Where the boundary is

One rule decides most of the design questions that follow:

> **A single cluster's invariants are enforced by the data plane. The control plane owns sequencing, direction and execution — and only *predicts* what the data plane will refuse.**

[...]

PR #477, opened 2026-08-06
Reject datastore table teardown while a URI-bound label references it

em3s:
## Summary

Disabling or dropping an HBase table only checked for a `StorageEntity` naming it. A label can bind straight to the table with a `datastore://{namespace}/{table}` URI and no storage entity at all, and every table the v3 API creates takes that form. So a table serving live reads and writes could be disabled or dropped.

Relates to #370

## Changes

- Add `DatastoreTableReferences`, resolving both binding forms
- Use it in `DatastoreHBaseService` and `HBaseController`, replacing three copies of the inline storage scan
- Refuse to answer when a metadata scan hits `metadataFetchLimit` — a truncated page can't prove a table is unreferenced
- Wrap the admin call in `Mono.defer` so the guard runs first. An argument to `then` is evaluated eagerly, which put connection setup ahead of the check

## How to Test

```
```

## Behavior change

Teardown that used to succeed is now rejected — a table bound by an active label has to be deactivated first. `/graph/v2/admin/hbase` stays cluster-blind as before, so for a remote cluster it errs toward refusing.

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/service/ddl/DatastoreTableReferences.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/service/ddl/DatastoreTableReferences.kt
@@ -1,0 +1,111 @@
+package com.kakao.actionbase.v2.engine.service.ddl
+
+import com.kakao.actionbase.engine.EngineConstants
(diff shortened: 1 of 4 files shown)


em3s:
Optimistic

PR #478, opened 2026-08-06
Expose what references a datastore table

em3s:
## Summary

Deciding whether an HBase table is safe to tear down means reconstructing the `alias -> label -> storage -> table` chain client-side. The server already resolves that chain to guard its own disable and drop, so let clients ask instead.

Stacked on #477. Relates to #370

## Changes

- `GET .../tables/{tableName}/references` for one table
- `GET /graph/v3/datastore/hbase/references` for the whole cluster from a single scan, keyed by `namespace:tableName`

Per-table resolution costs a metadata scan per database, so anything walking a cluster should use the bulk form. It sits outside `tables/` to avoid colliding with `tables/{tableName}`. Inactive bindings are included so an operator can tell a teardown in progress from a table nothing ever pointed at.

## How to Test

```
```

## AI Assistance


em3s:
Optimistic

PR #479, opened 2026-08-06
Test the HBase datastore controllers over HTTP

em3s:
## Summary

Nothing under `/graph/v3/datastore/hbase` was reachable from a test — the controllers only exist under `actionbase.datastore.type=HBASE` and the test profile runs on `memory`. This adds the harness and uses it to cover #477 and #478.

Stacked on #478. Relates to #370

## Changes

- `HBaseDatastoreE2ETestBase`: flips the datastore type and takes over the `HBaseAdmin` bean with a mock
- Add mockk to `:server` test deps — `HBaseAdmin` is final
- Cover both reference endpoints, plus the teardown guard on v3 and on the v2 devtools controller

Proving a rejection needs the admin call stubbed with a publisher that fails on subscribe rather than left unstubbed. The service builds the call before the guard runs — an argument to `then` is evaluated eagerly — so an unstubbed mock reports a failure that isn't one.

## How to Test

```
./gradlew :server:test --tests "*DatastoreTableReferencesE2ETest*" --tests "*HBaseAdminTeardownGuardE2ETest*"
```

## AI Assistance

### DIFF
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -59,6 +59,9 @@ dependencies {
 
     // test
     testImplementation(testFixtures(project(":core")))
+    // The datastore controllers depend on HBaseAdmin, a final class, so standing them up in a test
+    // context needs inline mocking.
+    testImplementation(Dependencies.Testing.MOCKK)
 }
 
 gitProperties {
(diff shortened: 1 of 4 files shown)


em3s:
Optimistic

PR #437, opened 2026-07-20
Introduce `queue/v1`: message queue over immutable edge tables

em3s:
## Summary

queue/v1 is a message queue that also serves as a refresh / delay queue. It is an **engine capability**: the runtime (enqueue / poll) lives in `engine` (`com.kakao.actionbase.engine.queue`) over the immutable edge table, with a thin `server/` HTTP layer and queue DDL (create / enable / disable / delete) over the v3 table API.

A queue *is* an `ImmutableEdge` table: `source` = partition, `target` = `id` (server-assigned ULID), a LONG `seq` property indexed **ASC** for per-partition order, and an opaque `value` (any JSON). Enqueue is a lock-free append; poll is a per-partition range scan over `seq` with a forward `offset` cursor and an optional `until` upper bound (refresh / delay).

Message = `{ key, seq, value }`; endpoints under `/queue/v1/namespaces/{ns}/queues`. Full design — data model, and how every endpoint lowers to a v3 edge scan / table DDL — is in the design comment below.

Scoped to admin + enqueue + poll. Eviction / retention is a **TBD follow-up** — a background scan-and-delete sweep over the `ImmutableEdge` (which forbids in-place update, not deletion) — tracked separately.

Closes #428

## Changes

- **engine** (`com.kakao.actionbase.engine.queue`) — `QueueService` (enqueue / poll), `PartitionHasher`, `QueueDescriptorCodec`, `QueueSchema`, and the runtime payloads. Enqueue appends via `MutationService.mutate` INSERT with `lock=false` and assigns the ULID `id`; poll is a single-partition `seq`-index scan (`seq:gt:offset` for a log, `seq:bt:offset+1,until` for refresh). Wired as a bean in `GraphConfiguration`.
- **server** (`api/queue/v1`) — `MessageController` (enqueue / poll) delegating to the engine; `api/queue/v1/metadata/QueueController` + `QueueMetadataService` build / inspect / toggle / delete the backing table over `V3CompatService`, with delete guarded (`409` unless inactive).

No changes to the core immutable-edge binding — the queue reuses it as-is.

## How to Test

- `./gradlew :engine:test --tests "com.kakao.actionbase.engine.queue.*"` — `PartitionHasher`, descriptor codec.
- `./gradlew :server:test --tests "*queue.v1.*"` — E2E (admin lifecycle incl. enable / disable / delete-guard, enqueue with ULID `id`, poll order / pagination / `until`).
- `./gradlew :engine:spotlessCheck :server:spotlessCheck`.

[...]

PR #481, opened 2026-08-06
Require a token on the queue api

em3s:
## Summary

A queue *is* a v3 edge table — `queue/v1` lowers to edge scans and table DDL over one. But `TokenAuthenticationFilter` only ever guarded `/graph/v2` and `/graph/v3`, so `/queue/v1` was an unauthenticated way to the same data the token is there to protect:

- `POST`/`DELETE .../queues` — create and delete queues, which is v3 table DDL
- `PUT .../enable`, `.../disable` — stop a queue
- `POST .../messages` — inject messages
- `GET .../partitions/{p}/poll` — read another tenant's contents
- `DELETE .../partitions/{p}/messages?offset=` — commit their offsets away

This was drift, not a decision. `protectedPaths` had not been touched since the initial open-source release; `queue/v1` arrived later in #437 and nobody revisited the list. The same gap let `/control` through until #480. Note that `/queue/v1` is already in the read-only filter's prefixes — the codebase treats it as a mutating surface everywhere except here.

## Changes

- `TokenAuthenticationFilter.protectedPaths` is now `PathPrefixes.FILTERED`, every routable prefix
- `the queue api needs no token` becomes `the queue api needs a token`, plus a valid-token case

## ⚠️ Not code-only — read before merging

Any queue client that sends no `Authorization` header starts getting 401 the moment a deployment has `kc.graph.use-token` on. **Audit the queue clients and roll tokens out to them first.** The shipped default is `use-token: false`, so a deployment that never enabled tokens is unaffected either way.

Nothing in this repo calls `/queue/v1` outside tests, so the blast radius is entirely in callers we cannot see from here.

## How to Test

```
./gradlew :server:test --tests "*TokenAuthenticationFilterTest*"
```

The probes are covered too — `/graph/health`, `/graph/health/liveness` and `/` must keep answering unauthenticated or readiness fails on every pod.

## AI Assistance

### DIFF
--- a/server/src/main/kotlin/com/kakao/actionbase/server/filter/TokenAuthenticationFilter.kt
+++ b/server/src/main/kotlin/com/kakao/actionbase/server/filter/TokenAuthenticationFilter.kt
@@ -15,9 +15,9 @@ class TokenAuthenticationFilter(
 ) : WebFilter {
     private val log = LoggerFactory.getLogger(TokenAuthenticationFilter::class.java)
 
-    // Tokens cover the operational apis too, not just the data plane they reach. QUEUE is left out
-    // as it always has been - adding it would start rejecting clients that send no token.
-    private val protectedPaths = PathPrefixes.GRAPH + PathPrefixes.CONTROL
+    // Every routable prefix. A queue is a v3 edge table, so leaving it open left an unauthenticated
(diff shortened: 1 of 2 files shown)


em3s:
Optimistic

PR #482, opened 2026-08-06
Reach a tenant's clusters from the control plane

em3s:
## Summary

#480 resolved a tenant's clusters into a `Topology` and then stopped: `baseUrl(tenant, side)` and `sidesFor(tenant, fanout)` were written and tested with nothing calling them. This is the layer that calls them.

`ClusterClient` reaches the clusters of one tenant and returns one entry per side. **A failing side does not fail the call.** An operator asking what a tenant holds needs to know which cluster is unwell, and the healthy side's answer is still worth having — so each side comes back with either a body or its reason, in a stable order however slow either was. The cause is surfaced next to the message, because that is where the reason a connection failed actually lives.

The caller's `Authorization` and `Actor-ROLE` travel with the request: the cluster should see who is asking, not who is relaying.

Nothing mutates yet. Per the roadmap in #370, destructive execution lands after authorization.

## Changes

- Add `ClusterClient` and `SideResponse` under `server/control/cluster`
- Add `actionbase.control.request-timeout` (default 10s) — how long one cluster may take before its side is reported as failed
- Add `GET /control/tenants/{tenant}/datastore/tables?fanout=true`, merging each side's table listing

## How to Test

```
./gradlew :server:test --tests "*ClusterClientTest*" --tests "*ControlDatastoreE2ETest*"
```

There is no stub-server dependency in this repo and this PR does not add one: the unit tests substitute an exchange function for the clusters, so nothing binds a port. Covered there: per-side ordering, fanout on and off, a failing side not hiding a healthy one, a non-2xx side reporting its status, the timeout, and header forwarding.

`ControlDatastoreE2ETest` points the topology at a closed port, which is the interesting case over http — an unreachable cluster must be reported as *that side* failing with a 200 overall, not as the request failing.

## Notes

A `CONTROL` instance refuses `/graph/v2`, `/graph/v3` and `/queue/v1` on itself, so it cannot be its own data plane. That rules out the tempting shortcut of pointing a test's `active-url` at its own port.

## AI Assistance

### DIFF
--- a/server/src/main/kotlin/com/kakao/actionbase/server/api/control/datastore/ControlDatastoreController.kt
+++ b/server/src/main/kotlin/com/kakao/actionbase/server/api/control/datastore/ControlDatastoreController.kt
@@ -1,0 +1,75 @@
+package com.kakao.actionbase.server.api.control.datastore
+
+import com.kakao.actionbase.engine.datastore.hbase.admin.HBaseTableInfo
+import com.kakao.actionbase.server.configuration.ConditionalOnControlRole
+import com.kakao.actionbase.server.configuration.HttpHeaderConstants
+import com.kakao.actionbase.server.control.cluster.ClusterClient
+import com.kakao.actionbase.server.control.cluster.SideResponse
+import com.kakao.actionbase.server.control.topology.Side
+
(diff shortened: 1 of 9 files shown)


em3s:
Optimistic

PR #483, opened 2026-08-06
Answer what is left to clean up across the fleet

em3s:
> Builds on #482 — merge that first. Both commits show here because CI only runs on PRs based on `main`; the second commit is this PR's change, and the diff collapses to it once #482 lands.

## Summary

A cleanup client needs to know, per datastore table, what each side holds and what still binds to it. Every client derives that itself today — and two of the facts involved are ones no single cluster can supply.

`GET /control/htables` answers it once for the whole fleet.

**Sides are symmetric.** A table's state is one entry per side, so a table that survives only on the standby is an absent active side rather than a second kind of row. Clients model that as a separate list because they see one cluster at a time; the control plane does not have to.

**The chain is read, not reconstructed.** References come from the endpoint added in #478, and each one carries the side it was found on — a label still active on the standby is exactly the thing that blocks a teardown halfway through.

**Sharing is noticed without anyone declaring it.** Two tenants configured against the same cluster resolve to the same cluster key (`host:port`), so a table held by another tenant shows up as `sharedWith` — and it names the tenant, because "blocked" without "by whom" leaves the operator where they started.

**A failed side is reported as failed.** An empty table list from a cluster that never answered would read as *nothing left to clean up*, which is the one wrong answer this endpoint must never give.

## Changes

- Add `CleanupProjection` (pure) and `CleanupService` under `server/control/cleanup`
- Add `GET /control/htables` with optional `tenant` and `env` filters
- Add `Topology.cluster(tenant, side)` — the cluster key that makes sharing visible
- `Env.of` accepts the spelling config uses: `env: prod` in yaml and `?env=prod` in a url should not disagree, and an unknown value names the ones that exist
- Extract `CallerHeaders` so both control endpoints forward the caller's identity the same way

## How to Test

```
./gradlew :server:test --tests "*CleanupProjectionTest*" --tests "*ControlCleanupE2ETest*" --tests "*TopologyTest*"
```

`CleanupProjectionTest` drives the projection from fixtures — standby-only tables, a table held from another tenant, references carrying their side, and a failed side not reading as a clean fleet. `ControlCleanupE2ETest` points every configured cluster at a closed port, so the all-sides-failed path is exercised over http.

## AI Assistance

### DIFF
--- a/server/src/main/kotlin/com/kakao/actionbase/server/api/control/cleanup/ControlCleanupController.kt
+++ b/server/src/main/kotlin/com/kakao/actionbase/server/api/control/cleanup/ControlCleanupController.kt
@@ -1,0 +1,34 @@
+package com.kakao.actionbase.server.api.control.cleanup
+
+import com.kakao.actionbase.server.configuration.ConditionalOnControlRole
+import com.kakao.actionbase.server.configuration.HttpHeaderConstants
+import com.kakao.actionbase.server.control.cleanup.CleanupService
+import com.kakao.actionbase.server.control.cleanup.HtablesView
+import com.kakao.actionbase.server.control.cluster.CallerHeaders
+import com.kakao.actionbase.server.control.topology.Env
+
+import org.springframework.http.HttpHeaders
+import org.springframework.web.bind.annotation.GetMapping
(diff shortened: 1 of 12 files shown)


em3s:
Optimistic

Issue #470, opened 2026-07-24
Settle v3 vocabulary on `database`/`collection` (`table` reserved for HBase)


em3s:
## Context

Container terms piled up across generations: v2 `service`/`label`, v3 `database`/`table`, queue `namespace`/`queue`, datastore `datastore`/`namespace`.

`table` is the pain: the logical `table` and the physical HBase `table` both appear in the API (`/databases/{db}/tables/{t}` vs `/datastore/hbase/tables/{name}`), with no way to tell them apart.

## Decision

v3 logical hierarchy is `database` > `collection` > `edge`/`vertex`/`multi-edge`. Physical stays `datastore` > `namespace` > `table` (HBase). So `table` means physical only, `collection` means logical only.

- `collection`: what modern DBs converge on (MongoDB, Firestore, vector DBs; ArangoDB uses vertex/edge collections).
- Keep `database`: Mongo is exactly `database > collection`, and only `table` changes so churn stays small. (`domain` etc. considered, dropped — taste, not correctness.)

Fold queue into `/graph/v3`:
- `/queue/v1/namespaces/{ns}/queues/{q}` → `/graph/v3/databases/{db}/collections/{c}`
- A queue is already an `ImmutableEdge` table (`QueueMetadataService` builds it via the v3 table API; `QueueService` runs enqueue/poll as edge mutate/scan) — a surface move, not a storage migration.
- `partitions` nests under `messages` (`/messages/partitions/{p}/poll`), matching `/edges/scan/{index}`.
- `namespace`/`queue` container words disappear; `namespace` stays physical-only.

## Consequences

- Add `/collections/{collection}`; keep `/tables/{table}` as a maintained compat alias (same handler, internal code untouched). v3→v2 mapping unchanged (`database`→`service`, `collection`→`label`).
- Keep `/queue/v1` as a compat alias.
- Docs/CLI/responses lead with `collection`; `table` stays supported. Ambiguity closes when `tables` is retired.
- Follow-ups: body field naming (`table`/`collection`), CLI alias, Queue as a first-class model (#471).

em3s:
Closing as stale. We'll revisit if it comes up again.

PR #484, opened 2026-08-06
Plan cleanup work instead of letting clients derive it

em3s:
## Summary

The teardown order and its preconditions are the part every operational client reimplements: replication off before disable or the standby stops tracking, unlink before drop or the metadata dangles, delete the metadata only once the table is gone. Getting one wrong breaks a cluster, so the rule belongs in one place.

`POST /control/jobs` answers what an action would do, over a batch of targets. Nothing takes effect: `dryRun=false` is refused until authorization lands.

Relates to #370

## Three choices worth naming

**Targets are a list.** A per-table endpoint is what makes each client build its own execution queue, and then makes the next client build it again.

**`dryRun` is a flag on the endpoint that will execute, not a separate preview endpoint.** Two code paths computing the same plan is exactly how a plan and its execution come to disagree — and the plan an operator reviewed has to be the plan that runs.

**Refusals are typed.** An operator facing a disabled action needs to know which precondition to satisfy, and a test needs to assert the right one was named, so `Precondition` is an enum with the human detail beside it:

```json
{
  "target": {"tenant": "...", "table": "..."},
  "action": "DROP_TABLE",
  "ok": false,
  "refusal": "NEEDS_LOCAL_REPLICATION",
  "detail": "STANDBY still replicates: turn replication off first",
  "steps": []
}
```

These are **predictions**. The data plane enforces these invariants itself and stays the one that says no — #477 already refuses a teardown while a reference is live. Predicting them here is how a client learns before it asks, not a second source of truth.

## Rules encoded

| Action | Steps | Refused when |
| --- | --- | --- |
| `REPLICATION_OFF` / `REPLICATION_ON` | replication off / on | already there; `REPLICATION_ON` on a disabled table |
| `DISABLE_TABLE` / `ENABLE_TABLE` | disable / enable | already there; `DISABLE_TABLE` while replication is on |
| `UNLINK` | deactivate collections, then storages | nothing active still binds it |
| `DROP_TABLE` | drop, per side that has it | metadata still bound; still enabled; still replicating; already gone |
| `DELETE_METADATA` | delete collections, then storages | the table still exists anywhere |

Fanout is partial — only the sides not already at the target state get a step, so an active/standby pair that has drifted apart is brought together rather than refused. Metadata actions (`UNLINK`, `DELETE_METADATA`) stay on the active side, because metadata is not replicated. A table living only on the standby is still droppable there.

## Naming

Action and step names follow the #470 decision: `table` is the physical HBase table, `collection` the logical v3 entity. That matters most for the two steps that used to be `DROP_TABLE` and `DELETE_TABLE` — a verb apart, but one loses the rows and the other only the definition. They are now `DROP_TABLE` and `DELETE_COLLECTION`.

`CleanupAction.destructive` marks the same boundary in the type rather than in prose: `DROP_TABLE` and `DELETE_METADATA` cannot be undone, everything else is a staging step. That is the line a confirmation prompt belongs on.

## How to Test

```
./gradlew :server:test --tests "*CleanupPlannerTest*" --tests "*ControlJobE2ETest*"
```

Because the plan is dry, the grid can be walked exhaustively: every refusal test asserts the precondition it named, not just that it failed.

`POST /control/jobs` is the first operational write, so the endpoint classification gains a `CONTROL_WRITE_ENDPOINTS` set — refused by a read-only instance like any other write, but served by a control instance, which is why it cannot simply join the data-plane write list.

## Not touched

`CleanupAction` keeps its name even though `ENABLE_TABLE` and `REPLICATION_ON` undo rather than clean up. Renaming it reaches `CleanupPlan`, `CleanupPlanner` and `CleanupJobService`; better done alongside the executor.

## AI Assistance

### DIFF
--- a/server/src/main/kotlin/com/kakao/actionbase/server/api/control/cleanup/ControlJobController.kt
+++ b/server/src/main/kotlin/com/kakao/actionbase/server/api/control/cleanup/ControlJobController.kt
@@ -1,0 +1,36 @@
+package com.kakao.actionbase.server.api.control.cleanup
+
+import com.kakao.actionbase.server.configuration.ConditionalOnControlRole
+import com.kakao.actionbase.server.configuration.HttpHeaderConstants
+import com.kakao.actionbase.server.control.cleanup.CleanupJobService
+import com.kakao.actionbase.server.control.cleanup.JobRequest
+import com.kakao.actionbase.server.control.cleanup.JobView
+import com.kakao.actionbase.server.control.cluster.CallerHeaders
+
+import org.springframework.http.HttpHeaders
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestHeader
+import org.springframework.web.bind.annotation.RestController
+
+import reactor.core.publisher.Mono
+
+/**
+ * Cleanup work, planned. `dryRun=true` is the only mode for now and returns the steps the action
+ * would take, or the precondition that stops it.
+ *
+ * Execution will be this same endpoint with `dryRun=false`, so that the plan an operator reviewed is
+ * the plan that runs - two code paths computing the same thing is how they come to disagree.
+ */
+@RestController
+@ConditionalOnControlRole
+class ControlJobController(
+    private val jobService: CleanupJobService,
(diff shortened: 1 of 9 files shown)


em3s:
Optimistic

PR #486, opened 2026-08-10
Purge tombstoned metastore rows from the control plane

### LINKED ISSUE #485: Purge tombstoned metastore rows from the control plane
## Background

This is a JDBC problem and it dies with JDBC — Phase 4 of the consolidation drops MySQL entirely. Until then the metastore accumulates rows nothing can remove.

A metadata `DELETE` is a tombstone: `DdlService.delete` mutates with `EdgeOperation.DELETE` and `AbstractLabel.updateExistingHashEdge` writes the row back with `Active.INACTIVE`. Only `EdgeOperationStatus.PURGED` issues a real SQL `DELETE`, and metadata DDL never takes that path. Nothing filters them out on read either — `JdbcHashLabel.scanStorage` applies `LIMIT` before the active filter, which runs afterwards in `AbstractLabel.scan` over the decoded rows.

That already breaks things. `DatastoreTableReferences.whole()` refuses when a scan hits `metadataFetchLimit`, so once a service crosses 1000 tombstones, `GET /control/htables` and `POST /control/jobs` stop working for that tenant. Cleanup is the workflow that needs this data.

Relates to #347

Relates to #370

## Design

`active` is not a column. It is the first field of the binary value and `deleteTs` sits at a fixed property hash, so `AbstractEdgeEncoder.decodeHashEdgeValueFromBytes` reads both without the label's schema — `hashToFieldNameMap` only names user properties. Scoping runs off the primary key rather than a `LIKE` prefix, which would need the label id: `MetadataTable` is a `LongIdTable`, so `WHERE id > :cursor ORDER BY id LIMIT :batch` gives an indexed, resumable walk and the service filter applies to the decoded `src`.

```
POST /control/metastore/purge/candidates   query    → PurgeSet
POST /control/metastore/purge/execute      PurgeSet → outcome
POST /control/metastore/purge/restore      PurgeSet → outcome

PurgeSet {
  metastore, table, service, generatedAt, scanned, nextCursor,
  rows: [ { k, v, createdAt, createdBy, modifiedAt, modifiedBy, updateTs } ],
  undecodable: [ { k, reason } ]
}
```

**Stateless, one document.** What `candidates` returns is exactly what the other two accept, so the CLI saves a response and later posts it back without transforming anything. The contents therefore reach the client before anything is deleted, which makes the candidate file the backup — there is no window where the server deleted rows whose backup was lost in a dropped response. Commit is the operator keeping the file; revert is feeding it to `restore`.

**The target is a MySQL table, not a tenant.** This does not talk to a cluster, so a tenant name would stand in front of what it actually opens. Credentials stay in a configured list of openable targets and a target absent from it is refused, or the endpoint would connect wherever a request body pointed it. There is no `side` either — active and standby share one metastore, so purging once is purging both.

**A row is its `k` and `v`.** `execute` issues `DELETE WHERE k = :k AND v = :v`, so it deletes exactly the data it was shown: a resurrected tombstone has a different `v` and survives, and a row already removed matches nothing, which makes a retry harmless. `restore` inserts only where the key is absent.

A row must be `active=false` and its `deleteTs` older than the requested window. A row that fails to decode is never touched. A service scope is required, and every batch has a row cap.

```
actionbase purge plan --metastore wish --service wish-prod --older-than 30d --max 500 -o wish-01.json
actionbase purge show|apply|restore wish-01.json
```

The control plane holding a direct database tool departs from the boundary rule in #370. I'm accepting it because the tool is temporary and its target is a table the data plane is already abandoning. It lives under `server/control/metastore/` and touches no data-plane path.

## Task

- [ ] Add `actionbase.control.metastores.*` — the JDBC URI and table pairs the purge may open, credentials from the environment, resolved at startup
- [ ] Add the PK-cursor scanner: decode `active` and `deleteTs` with an empty field-name map, filter by service, report undecodable keys
- [ ] Add `PurgeSet` as one type, serving as the `candidates` response and the request body of the other two
- [ ] Add the three endpoints, rejecting a `PurgeSet` whose target is not in the configured list
- [ ] Classify them in `ReadOnlyRequestFilter` — two are writes and belong in `CONTROL_WRITE_ENDPOINTS`
- [ ] Add `purge plan` / `show` / `apply` / `restore` to the Go CLI
- [ ] E2E test that posts the `candidates` response back verbatim, covering a resurrected tombstone, an undecodable row, a repeated `execute`, and a restore over a live key

## Done When

An operator can list, delete and restore one service's tombstones entirely over the API and through the CLI. Only the listed rows are deleted, byte for byte; a row that cannot be decoded never is; repeating `execute` changes nothing.

## Notes

The root cause outlives this issue: `scanStorage` cannot filter on `active` in SQL. An `active` column with an index would fix it properly, at the cost of a schema change and a backfill.

`GET /control/htables` and `POST /control/jobs` should start working again for tenants that were over the limit. That is the check that this was worth doing.

em3s:
## Summary

The metastore keeps every row it was ever told to delete. A metadata `DELETE` writes the row back with `Active.INACTIVE`, and `JdbcHashLabel.scanStorage` applies `LIMIT` before the active filter, so tombstones spend the budget a live scan needs. Past 1000 of them under one service, `DatastoreTableReferences.whole()` refuses and the control plane's own cleanup endpoints stop answering.

This adds a purge that reads a metastore table directly, hands back what it would delete, deletes exactly that, and can put it back. Stateless: `candidates` returns a `PurgeSet`, and `execute` and `restore` accept the same document unchanged, so the file an operator holds is both the plan and the backup.

Temporary. It targets a table the data plane is already abandoning and goes away with JDBC in Phase 4 of #347.

Closes #485

## Plan

Created by **claude code (opus 5)**

- [x] 1. `actionbase.control.metastores.*` config — the JDBC URI and table pairs the purge may open, credentials from the environment, resolved at startup
- [x] 2. PK-cursor scanner — decode `active` and `deleteTs` with an empty field-name map, filter by service, report undecodable keys
- [x] 3. `PurgeSet` as one type, serving as the `candidates` response and the request body of the other two
- [x] 4. The three endpoints, rejecting a `PurgeSet` whose target is not in the configured list
- [x] 5. Endpoint classification in `ReadOnlyRequestFilter` — `execute` and `restore` are writes
- [x] 6. `purge plan` / `show` / `apply` / `restore` in the Go CLI
- [x] 7. E2E test posting the `candidates` response back verbatim — resurrected tombstone, undecodable row, repeated `execute`, restore over a live key

## Progress

- [x] 1 — **claude code (opus 5)** (commit 5cfebd5)
- [x] 2 — **claude code (opus 5)** (commit 8fe7c64)
- [x] 3–7 — **claude code (opus 5)** (commit a02730f)
- [x] Review round 1 fixes — **claude code (opus 5)** (commit db3c10d)
- [x] Tests for the round 1 fixes, and the `resolve` comment corrected — **claude code (opus 5)** (commit 0fcd65f)

## Changes

**engine** — `metastore/purge/`, standalone. `MetastorePurge` walks `WHERE id > :cursor ORDER BY id`, decodes two fields, and deletes on `k` and `v` together. `PurgeFields` reads the active flag off the front of the encoded value and the service off the front of the encoded key. `PurgeTable` is its own Exposed definition.

**server** — `control/metastore/` holds the allowlist, the credentials and the service; `api/control/metastore/` holds the three endpoints. `candidates` joins the read-only filter's allowed suffixes, `execute` and `restore` join `CONTROL_WRITE_ENDPOINTS`.

**cli** — `purge plan | show | apply | restore`, driving the endpoints over a file. `apply` and `restore` need `--yes`.

## Three things worth naming

**Nothing from the data model.** The purge reads `active` and the service straight off the encodings with `OrderedBytes` and `ValueUtils`. No `HashEdgeValue`, no `DecodedEdge`, no `EdgeSchema`, no `Graph`. A tool that dies with JDBC should not be entangled with what is being retired around it, and it means no new dependency in `server` at all.

**How old a tombstone is comes from `update_ts`, not the encoded `delete_ts`.** Reaching that property means walking the whole property list, while the column says the same thing for a row that is currently inactive — the last write was the one that deactivated it — and says it in a `WHERE` clause against an index.

**One document, three endpoints.** `candidates` returns it and the other two take it back unchanged, so the CLI never reshapes anything. The rows therefore reach the client before anything is deleted, which is what makes the saved file a backup rather than a convenience. Committing a purge is keeping the file; reverting one is posting it to `restore`.

## How to Test

```
./gradlew :engine:test --tests "*MetastorePurge*"
./gradlew :server:test --tests "*Metastore*" --tests "*ControlPurge*" --tests "*RequestFilterTest*"
cd cli && go test ./...
```

`ControlPurgeE2ETest` runs against a real table and posts the `candidates` response back byte for byte, which is the claim that has to keep holding.

## Not touched

`JdbcHashLabel.cad` is broken — its `deleteWhere` lambda has `k eq ...` and `v eq ...` as separate statements, so Exposed takes the last as the predicate and the delete matches on value alone, ignoring the key. `deleteOnLock` is the caller. Out of scope here and it needs its own issue.

The root cause outlives this PR: `scanStorage` still cannot filter on `active` in SQL. An `active` column with an index would fix it properly, at the cost of a schema change and a backfill.

## AI Assistance

### DIFF
--- a/cli/internal/command/command_type.go
+++ b/cli/internal/command/command_type.go
@@ -85,6 +85,11 @@ var (
 		name:    "migrate",
 		command: "migrate <plan [-o <file>]|apply [-i <file>]>",
 	}
+
+	TypePurge = Type{
+		name:    "purge",
+		command: "purge <plan --metastore <name> --service <service> [--older-than <days>] [--max <rows>] [--file <file>]|show [--file <file>]|apply [--file <file>] --yes|restore [--file <file>] --yes>",
+	}
 )
 
 func (t Type) GetName() string {
--- a/cli/internal/command/purge.go
+++ b/cli/internal/command/purge.go
@@ -1,0 +1,229 @@
+package command
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/kakao/actionbase/internal/client"
+	"github.com/kakao/actionbase/internal/command/model"
+	"github.com/kakao/actionbase/internal/util"
+)
+
+const (
+	defaultPurgeFile = "purge.json"
+	candidatesPath   = "/control/metastore/purge/candidates"
+	executePath      = "/control/metastore/purge/execute"
+	restorePath      = "/control/metastore/purge/restore"
+	defaultOlderThan = 30
+	defaultMaxRows   = 500
+	purgeFilePerm    = 0o600
+	httpOK           = 200
+)
+
+// purgeSet is only enough of the document to summarise it. The file is posted
+// back byte for byte, so nothing here reshapes it - a field the server adds
+// later still round-trips even though this struct has never heard of it.
+type purgeSet struct {
+	Metastore   string `json:"metastore"`
+	Table       string `json:"table"`
+	Service     string `json:"service"`
+	GeneratedAt string `json:"generatedAt"`
+	Scanned     int    `json:"scanned"`
+	NextCursor  *int64 `json:"nextCursor"`
+	Rows        []struct {
+		K string `json:"k"`
+	} `json:"rows"`
+	Undecodable []struct {
+		K      string `json:"k"`
+		Reason string `json:"reason"`
+	} `json:"undecodable"`
+}
+
+type purgeResult struct {
+	Requested int `json:"requested"`
+	Applied   int `json:"applied"`
+	Skipped   []struct {
+		K      string `json:"k"`
+		Reason string `json:"reason"`
+	} `json:"skipped"`
+}
+
+// Purge drives the three metastore purge endpoints.
+//
+// The file written by plan is the response as it arrived, and apply and restore
+// post it back unchanged. That is what makes it a backup: the rows are on disk
+// before anything is deleted, so a lost response can never leave rows gone with
+// no copy of them. Committing a purge is keeping the file; reverting one is
+// handing the same file to restore.
+type Purge struct {
+	client *client.ActionbaseClient
+}
+
+func NewPurge(c *client.ActionbaseClient) *Purge {
+	return &Purge{client: c}
+}
+
+func (p *Purge) Execute(args []string) *model.Response {
+	if len(args) < 1 {
+		return model.Fail(fmt.Sprintf("Usage: %s", p.GetType().GetCommand()))
+	}
+	parser := util.ParseArgs(args[1:])
+
+	switch args[0] {
+	case "plan":
+		return p.plan(parser)
+	case "show":
+		return p.show(fileArg(parser))
+	case "apply":
+		return p.send(executePath, "delete", parser)
+	case "restore":
+		return p.send(restorePath, "restore", parser)
+	default:
+		return model.Fail(fmt.Sprintf("Usage: %s", p.GetType().GetCommand()))
+	}
+}
+
+func (p *Purge) plan(parser *util.Parser) *model.Response {
+	metastore, ok := parser.Get("metastore")
+	if !ok {
+		return model.Fail("--metastore is required: it names a metastore configured on the control plane")
+	}
+	service, ok := parser.Get("service")
+	if !ok {
+		return model.Fail("--service is required: a purge is always scoped to one service")
+	}
+
+	status, body := p.client.PostRaw(candidatesPath, map[string]any{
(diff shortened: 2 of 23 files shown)


em3s:
stale

PR #418, opened 2026-07-15
Remove the dead MetastoreInspector and its dump endpoints

em3s:
## Summary

Remove `MetastoreInspector` and `MetastoreController`, which only powered the debug `/graph/v2/metastore/{global,local}` dump endpoints. Nothing else consumes them.

First of a 3-PR stack that reworks the local metastore backend:
1. **This PR** — drop the dead inspector/endpoints.
2. #TBD_A — add `ByteArrayStore`-backed Hash/Indexed labels.
3. #413 — back the local metastore with the in-memory `ByteArrayStore`.

## Changes

- Delete `MetastoreInspector` and `MetastoreController`.
- Drop the `metastoreInspector` / `localMetastoreInspector` beans in `GraphConfiguration` and the `Graph.metastoreInspector` property.
- Remove the `/graph/v2/metastore/{global,local}` entries from the read-only endpoint allowlist test.
- Drop the now-obsolete `dumpMetastore` assertion from `InitialSpec`.

## How to Test

- `./gradlew build`
- `ReadOnlyRequestFilterTest` still passes its endpoint-exhaustiveness check with the two entries removed.

## AI Assistance

- [x] This PR was written largely with AI assistance.
  - Tool / model: Claude Opus 4.8

PR #487, opened 2026-08-10
Report what is left in the JDBC metastore before a metadata scan truncates

em3s:
## Summary

`JdbcHashLabel` fetches `graph.metadata-fetch-limit` rows per scan key and filters inactive ones out afterwards. A metadata `DELETE` only tombstones, so tombstones spend that budget, and past the limit the page is truncated silently — `AbstractLabel.scan` computes `hasNext` and `DdlService.getAll` drops it.

EdgeIndex on v3 is the fix. Until then I needed to see how close deployed servers are, so this adds a command that measures it.

`jdbc-metastore remaining` walks `/graph/v2/metastore/global` — the only endpoint that shows the rows a scan reads, tombstones included — and counts per `(src, labelId)`, the scan key `getAll` builds. Both values come off the server's decode, so nothing re-implements the codec. The endpoint is 0.4.x only, removed by #418, so the command probes first and says so.

No server change. Same tombstone problem as #485.

## Notes

- `--capacity` is required: it is the server's `graph.metadata-fetch-limit`, deployments override the OSS default of 1000, and actuator masks the value.
- Every window prints, grouped by namespace, since a metastore table can be shared between deployments.
- Bounded: counters per page, 500-row pages, five stop conditions (`last`, empty page, `--max-rows`, `--max-pages`, total reached). Whichever cap fires is named and exits non-zero. Ctrl-C reports what it counted.
- Not using the server's `prefix=` count: `buildQuery` builds `k LIKE '<prefix>%'` unescaped, and base64url keys contain `_`.

## Changes

- Add `jdbc-metastore remaining`, with `--capacity`, `--page-size`, `--max-pages`, `--max-rows`
- Add `GetStream` to the HTTP client, which keeps the real status code and decodes off the connection
- Add `DumpMetastore(page, size)` and the dump page model
- Register the command

## How to Test

```
cd cli && go test ./internal/command/ && go build ./... && go vet ./...
```

Against a 0.4.x server. Lower both to see the classification without a large table:

```
actionbase --host http://<0.4.x-host>:8080 -e "jdbc-metastore remaining --capacity 3 --page-size 2"
```

```
20 rows, capacity 3/window, pages of 2 (max 2000). Ctrl-C reports partial.
alpha         14 rows  4 windows
  service      OVER by 2   alpha:origin               5/3     live 5      tomb 0
  storage      OVER by 2   alpha:origin               5/3     live 1      tomb 4
  label        1 left      alpha:talk                 2/3     live 0      tomb 2
               1 left      alpha:wish                 2/3     live 0      tomb 2
beta           5 rows  2 windows
  service      2 left      beta:origin                1/3     live 1      tomb 0
  label        OVER by 1   beta:gift                  4/3     live 2      tomb 2
1 undecodable (window unknown, e.g. k="not-an-encoded-key")
20/20 rows, 6 windows, 2 namespaces. 3 over, 0 near. An over-capacity window is already truncated and does not say so.
```

All of these exit non-zero, as does a window over capacity:

```
4/20 rows, 1 window, 1 namespace. Stopped at --max-pages 2 - incomplete, 1 over is a lower bound.
--capacity is required: the server's graph.metadata-fetch-limit, which it does not expose (actuator masks it). OSS default 1000.
No /graph/v2/metastore/global on this server (0.4.x only), so this cannot be measured here.
Unauthorized. Set --authKey or ACT_API_KEY.
```

Sample output and the last two paths came from a stub server; the rest ran against real 0.4.x servers.

## Not touched

The root cause, which EdgeIndex addresses. Also `MetastoreInspector.buildQuery`'s `_` escaping, which is on 0.4.x and below rather than `main`, and `CheckConnection` exiting 0 on an unreachable host, which predates this and affects every command.

## AI Assistance

### DIFF
--- a/cli/internal/client/actionbase_client.go
+++ b/cli/internal/client/actionbase_client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"net/url"
+	"strconv"
 
 	"github.com/kakao/actionbase/internal/client/model"
 )
@@ -120,4 +121,23 @@ func (a *ActionbaseClient) PostRaw(path string, body any) (int, string) {
 
 func (a *ActionbaseClient) GetHost() string {
 	return a.client.baseUrl
+}
+
+// MetastoreDumpPath is served by 0.4.x and was removed on later servers, so a caller has to be
+// ready for a 404 rather than assume it is there.
+const MetastoreDumpPath = "/graph/v2/metastore/global"
+
+// DumpMetastore reads one page of the metastore, ordered by primary key.
+//
+// Ordered by id rather than k: it is the primary key, so the server's OFFSET pagination walks an
+// index it already has. Sorting by k would order by a base64 string for no benefit here.
+func (a *ActionbaseClient) DumpMetastore(page, size int) *Response[model.MetastoreDumpPage] {
+	u := &url.URL{Path: MetastoreDumpPath}
+	q := u.Query()
+	q.Set("page", strconv.Itoa(page))
+	q.Set("size", strconv.Itoa(size))
+	q.Set("sort", "id,asc")
+	u.RawQuery = q.Encode()
+
+	return GetStream[model.MetastoreDumpPage](a.client, u.String())
 }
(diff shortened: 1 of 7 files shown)


em3s:
Optimistic

PR #488, opened 2026-08-11
Serve the JDBC metastore's rows again, for as long as it exists

em3s:
## Summary

#418 removed `MetastoreInspector` and its dump endpoints as dead code, which they were. Something calls them now: `jdbc-metastore remaining` (#487) measures how much room each metadata scan window has left, and this is the only view of the rows such a scan actually reads. A `DELETE` rewrites its row with `Active.INACTIVE` instead of removing it, and `scanStorage` applies its `LIMIT` before anything filters on active, so every path above this one hides exactly what needs measuring.

Scaffolding for the move to EdgeIndex. It exists to see where deployments stand until the JDBC metastore is gone, and goes with it — hence `JdbcMetastore*` rather than a name claiming the general concept.

Narrower than what was removed:

- No `prefix`. It built `k LIKE '<prefix>%'` unescaped over base64url keys, whose `_` is a single-character LIKE wildcard, so roughly a third of prefixes matched other windows' rows.
- No `sort`. `id ASC` is the primary key, so paging walks an index that already exists. The parameter is accepted and ignored.
- No `/local`. No backing store for it on `main`.
- The row is `id`, `k` and four decoded fields. `v`, the property map, timestamp, direction and type are dropped.

The request shape still matches 0.4.x, so one client reads both. Blocking JDBC work goes to `boundedElastic`, which the removed version did not do.

## Changes

- Add `JdbcMetastoreInspector` with `dump(limit, offset)` and `count()`
- Add `JdbcMetastoreController` serving `GET /graph/v2/metastore/global`
- Register the bean and re-allow the path in the read-only filter

## How to Test

```
./gradlew build
./gradlew :server:bootRun --args='--spring.profiles.active=local'
curl -XPOST localhost:8080/graph/v2/service/demo -H 'Content-Type: application/json' -d '{"desc":"demo"}'
curl 'localhost:8080/graph/v2/metastore/global?page=0&size=3&sort=id,asc'
```

```json
{"id": 2, "k": "d2--KjRsb2NhbDpvcmlnaW4AK-m6YsMpfQ:3464656d6f00",
 "decoded": {"active": true, "src": "local:origin", "tgt": "demo", "labelId": 1773822659}}
```

Deactivate then delete that service: the row stays, with `active: false`, and the table is no shorter. That is the premise, and a test pins it.

The CLI reads it unchanged:

```
actionbase --host http://localhost:8080 -e "jdbc-metastore remaining --capacity 1000"
```

```
2 rows, capacity 1000/window, pages of 500 (max 2000). Ctrl-C reports partial.
local          2 rows  1 window
  service      998 left    local:origin               2/1000  live 2      tomb 0
2/2 rows, 1 window, 1 namespace. 0 over, 0 near.
```

## Not touched

The `COUNT(*)` per page, which keeps the response a Spring `Page` and compatible with 0.4.x. The table is small enough that it does not matter.

The root cause — `scanStorage` still cannot filter on `active` in SQL, and `getAll` still discards `hasNext`. EdgeIndex is where that gets fixed.

## AI Assistance

### DIFF
--- a/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/metastore/JdbcMetastoreInspector.kt
+++ b/engine/src/main/kotlin/com/kakao/actionbase/v2/engine/metastore/JdbcMetastoreInspector.kt
@@ -1,0 +1,88 @@
+package com.kakao.actionbase.v2.engine.metastore
+
+import com.kakao.actionbase.v2.core.code.DecodedEdge
+import com.kakao.actionbase.v2.core.code.KeyValue
+import com.kakao.actionbase.v2.engine.GraphDefaults
+import com.kakao.actionbase.v2.engine.storage.jdbc.MetadataTable
+
+import org.jetbrains.exposed.sql.Database
+import org.jetbrains.exposed.sql.SortOrder
+import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.transactions.transaction
+
+import reactor.core.publisher.Mono
+import reactor.core.scheduler.Schedulers
+
+/**
+ * Reads the JDBC metastore's rows as they sit on disk, tombstones included. No read path above this
+ * one shows them, and a scan key crowded with tombstones truncates silently.
+ */
+class JdbcMetastoreInspector(
+    private val metastore: Database,
+    private val table: MetadataTable,
+) {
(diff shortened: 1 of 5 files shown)


em3s:
Optimistic

PR #489, opened 2026-08-11
chore: mask metastore credentials in GraphConfig log output

eazyhozy:
## Summary

`Graph` logs the whole `GraphConfig` at INFO on startup, and the generated data class `toString()` renders every field — so the metastore user and password land in the log in plain text.

Before:

```
graph initialized with config: GraphConfig(phase=local, tenant=local, metastoreUrl=jdbc:mysql://db.example.com:3306/actionbase, metastoreUser=graph_app, metastorePassword=s3cr3t-p4ssw0rd, metastoreDriver=null, metastoreTable=ab_default, ...)
```

After:

```
graph initialized with config: GraphConfig(phase=local, tenant=local, metastoreUrl=jdbc:mysql://db.example.com:3306/actionbase, metastoreUser=****, metastorePassword=****, metastoreDriver=null, metastoreTable=ab_default, ...)
```

## Changes

- Add `GraphConfig.redacted()` — a `copy()` with the two credentials masked — and call it from the three sites that log the config.
- An empty credential stays empty rather than `****`, so "not configured" stays distinguishable from "configured".
- `metastoreUrl` is left as is: credentials reach the driver through the separate user/password fields, and the URL is the most useful part of the line when diagnosing which metastore an instance is talking to.

`copy()` keeps the generated `toString()`, so nothing has to be maintained by hand as the class grows. The trade-off is that `toString()` itself still renders the credentials, so a caller that logs the config directly would leak them again.

## How to Test

```bash
./gradlew :engine:compileKotlin :engine:compileTestKotlin
```

No test included: `redacted()` is a single `copy()` call, so a test would restate the implementation rather than constrain behaviour.

## AI Assistance


em3s:
LGTM. Optimistic merge.

Issue #490, opened 2026-08-12
Rebuild prepared queries on Actionbase Query


zipdoki:
## Background

#467 retired the stored query entity and its CRUD API because prepared queries were going to be redefined on Actionbase Query, and #466 said that redesign would be tracked separately. It never was, so I'm opening this to track it.

What I want is a query whose steps are fixed and whose values are not: a caller sends values alone, and whatever is derived from the body — a `transform`'s compiled plan above all — is built once and reused. That only pays off if the registered body never changes between calls, so where the values go is most of the design.

## Task

- [x] Take `transform` and `{name}` binding on Actionbase Query
   - #499
- [x] Add the prepared-query API skeleton — endpoints, payloads, service declaration 
   - #491
- [x] Store registered queries and the names pointing at them
- [x] Serve registration, reads, and execution by id or by name
- [x] Run an unregistered body with values sent alongside it (`POST /graph/v3/query`)
   - #500

## Done When

A query is registered once and called by a name with values alone, and the transform's plan is compiled on the first call rather than on each one. I can edit a registered query, or move a name onto a different one, without resending the parts that did not change.

## Notes

Two things need deciding, and neither belongs in a PR review.

Storage costs two system labels — one for queries, one for the names. #467 narrowed the metastore surface and #347's production run is still pending, so I'd want the owner of #347 to agree before this lands.

A name resolves through a cache held for `metastoreReloadInterval`, so after I move a name, instances disagree for up to a minute. That's the same bargain the label registry makes. If repointing a name is meant to be an incident response, a minute is too long, and the alternative is the metadata sync already used for labels.

One thing to record while I'm here: `ActionbaseQuery.stats` is declared but the v3 execution path ignores it. I store it and echo it back so nothing is silently dropped, but it does nothing yet.

zipdoki:
@em3s 
Opened #491 for the second item — the API surface only, nothing behind it yet. Worth a look before the storage steps land, since everything else hangs off the payload shape.

em3s:
Closing as stale.

PR #491, opened 2026-08-12
Add the prepared-query API skeleton (Step 2 of #490)

zipdoki:
## Summary

Step 2 of #490. This adds the API surface for prepared queries — endpoints, payloads, and a service declaration — with no implementation behind it. Every service method is `TODO("Not yet implemented")`, so what I'm asking for is a review of the shape.

Part of #490.

### Design notes

Three decisions I'd like checked:

- A query is registered without a name and gets an id back; a name points at that id separately. That's what lets a new query be checked by its id before it takes traffic, and moved back in one write.
- An argument declares a type. The body is stored as JSON, so `"limit": "{limit}"` is text where an `Int` belongs — the declared type is what makes it a number on the way in.
- These paths sit under `/graph/v3`, not under a database. A multi-hop query names its own database per step and may name several.

## Changes

- Add `PreparedQueryService` with `register`/`amend`/`get`/`list`/`delete`, the alias pair, and `query` — all `TODO`
- Add `PreparedQueryController` for `/graph/v3/prepared-queries` and `/graph/v3/prepared-queries/aliases`
- Take `arguments` on `POST /graph/v3/query/{id}`
- Allow the new read paths through `ReadOnlyRequestFilter`; `POST /graph/v3/query/{id}` reads, and its path ends in a variable so no suffix matches

## How to Test

```bash
./gradlew build
```

The endpoint-exhaustiveness check in `ReadOnlyRequestFilterTest` covers the ten new mappings. Nothing else can be tested until the service has a body, so the requests below are the shape under review rather than a transcript.

Register a query, then give it a name:

```bash
curl -X POST 'localhost:8080/graph/v3/prepared-queries' \
-H 'Content-Type: application/json' \
-d '{
  "comment": "top purchased items for one user, in one category",
  "arguments": [
    {"name": "entity", "type": "string", "comment": "the user asked about"},
    {"name": "category", "type": "string", "comment": "which ranking to read"},
    {"name": "limit", "type": "int", "comment": "how many ranked rows"},
    {"name": "minMetric", "type": "long", "comment": "drop anything below this"}
  ],
  "fetch": [
    {
      "type": "TOPK",
      "name": "ranked",
      "database": "shop",
      "table": "purchases_table",
      "topk": "top_purchased",
      "entity": {"type": "VALUE", "value": ["{entity}"]},
      "dimensionValues": {"category": "{category}"},
      "limit": "{limit}"
    }
  ],
  "transform": [
    {
      "type": "SQL",
      "name": "result",
      "sql": "SELECT target AS itemId, metric FROM ranked WHERE metric >= {minMetric} ORDER BY metric DESC"
    }
  ]
}'

curl -X POST 'localhost:8080/graph/v3/prepared-queries/aliases' \
-H 'Content-Type: application/json' \
-d '{"alias": "top_grocery", "target": "<id>", "comment": "home screen recommendations"}'
```

Run it, by the name or by the id:

```bash
curl -X POST 'localhost:8080/graph/v3/query/top_grocery' \
-H 'Content-Type: application/json' \
-d '{"arguments": {"entity": "user1", "category": "grocery", "limit": 100, "minMetric": 2}}'
```

```json
{
  "items": [
    {"name": "result", "data": [{"itemId": "item1", "metric": 2}], "rows": 1, "stats": [], "offset": null, "hasNext": false}
  ]
}
```

Declared arguments are all required, and a value of the wrong JSON type is read as the declared one, so `"limit": "100"` binds the same way `100` does.

<details>
<summary>Reads, moving a name, and editing a query in place</summary>

```bash
curl 'localhost:8080/graph/v3/prepared-queries?status=ACTIVE'
curl 'localhost:8080/graph/v3/prepared-queries/<id>'
curl 'localhost:8080/graph/v3/prepared-queries/top_grocery'
curl 'localhost:8080/graph/v3/prepared-queries/aliases'
curl 'localhost:8080/graph/v3/prepared-queries/aliases/top_grocery'
```

`{id}` takes either form. A name has to satisfy the metadata naming policy and a generated id can't, so the two can never collide and the shape of the path variable says which was meant. Listings answer with a bare array and take `?status=ACTIVE|INACTIVE|ALL`, as the table and alias listings do.

A registered query reads back as it was written, placeholders and all — what's stored is a template, not a query with values in it:

```json
{
  "tenant": "local",
  "id": "3f2c8a91-7d4e-4b12-9c33-0a51e6b8d204",
  "arguments": [
    {"name": "entity", "type": "string", "comment": "the user asked about", "nullable": false},
    {"name": "limit", "type": "int", "comment": "how many ranked rows", "nullable": false}
  ],
  "fetch": [{"type": "TOPK", "name": "ranked", "database": "shop", "table": "purchases_table", "topk": "top_purchased", "entity": {"type": "VALUE", "value": ["{entity}"]}, "limit": "{limit}", "...": "..."}],
  "transform": [{"type": "SQL", "name": "result", "sql": "SELECT target AS itemId, metric FROM ranked WHERE metric >= {minMetric} ORDER BY metric DESC"}],
  "stats": [],
  "active": true,
  "comment": "top purchased items for one user, in one category"
}
```

The response also carries the `revision`/`createdAt`/`createdBy`/`updatedAt`/`updatedBy` fields every other v3 descriptor does, at their defaults.

Moving a name is one call, and sending the old id back is the rollback:

```bash
curl -X PUT 'localhost:8080/graph/v3/prepared-queries/aliases/top_grocery' \
-H 'Content-Type: application/json' \
-d '{"target": "<other-id>"}'
```

Editing a query leaves the parts that weren't sent as they were:

```bash
curl -X PUT 'localhost:8080/graph/v3/prepared-queries/<id>' \
-H 'Content-Type: application/json' \
-d '{"transform": [{"type": "SQL", "name": "result", "sql": "SELECT target AS itemId FROM ranked WHERE metric >= {minMetric}"}]}'
```

</details>

## Not touched

Storage, and the two decisions recorded in #490 — where these rows live, and how fast a moved name propagates. `POST /graph/v3/query` also still takes a query with no `arguments`: that needs binding which doesn't exist yet, and declaring the field before it works would be a lie.

## AI Assistance


zipdoki:
Implementation goes on top of this as-is. Whatever it disagrees with gets fixed in a follow-up.

PR #493, opened 2026-08-19
Declare a top-K aggregation on a group and list what declares one (Steps 1 and 2 of #386)

zipdoki:
## Summary

Introduce top-K aggregation metadata on `Group` and expose a read endpoint listing tables that declare one.

A `Group` can now carry an `Aggregations` block. Today only `topk` is defined; the schema is designed so future aggregation kinds (e.g. HLL) can land as sibling fields without restructuring the payload.

`GET /graph/v3/aggregations` returns a flat list of `(type, database, table)` tuples for every table that declares an aggregation, filterable by `?type=`. The aggregate/mutate flow that consumes this metadata is out of scope for this PR and will land separately.

### Metadata example

A group inside a label creation payload:

```json
{
  "groups": [
    {
      "group": "purchased_count",
      "type": "COUNT",
      "fields": [{"name": "_target"}],
      "directionType": "OUT",
      "aggregations": {
        "topk": [
          {
            "topk": "top_purchased",
            "entity": "_source",
            "ranges": "_target:eq:{_target}",
            "dimension": "_target",
            "refreshAfterMillis": 31536000000,
            "rank": "commerce.purchases__topk"
          }
        ]
      }
    }
  ]
}
```

### Endpoint example

`GET /graph/v3/aggregations`

```json
{
  "items": [
    { "type": "TOPK", "database": "commerce", "table": "purchases" }
  ]
}
```

`GET /graph/v3/aggregations?type=TOPK` filters to a single aggregation type.

## Test plan

- `./gradlew :core:test --tests 'com.kakao.actionbase.core.metadata.common.AggregationsTest'`
- `./gradlew :engine:test --tests 'com.kakao.actionbase.engine.service.AggregationServiceSpec'`
- `./gradlew :engine:test --tests 'com.kakao.actionbase.v2.engine.entity.LabelEntityFunctionsSpec'`
- `./gradlew :server:test --tests 'com.kakao.actionbase.server.api.graph.v3.metadata.MetadataAggControllerE2ETest'`

## AI Assistance

- [x] This PR was written largely with AI assistance.
  - Tool / model: Claude Code (Opus 4.7)

Issue #386, opened 2026-06-19
Support per-entity Top-k (Shape 3 & 4)


zipdoki:
## Background

Implements Shape 3 and Shape 4 of the universal Top-K query engine defined in #369:

- Shape 3 topk(Target | Filter) — top-k targets filtered by a specific entity, all time
- Shape 4 topk(Target | Filter, Window) — top-k targets filtered by entity within a sliding time window

The same machinery covers global, non per-entity rankings by declaring `__GLOBAL__` as the entity value, so this is no longer strictly per-entity.

### Metadata

Top-k is declared under `aggregations` on an edge group definition, and a group can declare more than one.

```json
"aggregations": {
  "topk": [
    {
      "topk": "<topk_name>",
      "entity": "<entity_field>",
      "dimension": "<dimension_field>",
      "rank": "<database>.<table>",
      "ranges": "<range_expression>",
      "refreshAfterMillis": 3600000,
      "additionalProperties": ["<property>"]
    }
  ]
}
```

- `entity` — the field the ranking is scoped to. `__GLOBAL__` selects a single global ranking instead.
- `dimension` — the field being ranked. Every other non-bucket field becomes part of the dimension key, via `Group.dimensionFields`.
- `rank` — fully qualified name of the rank table, split by `parseFqn` into database and table.
- `ranges` — range expression applied when scanning the source group. It is interpolated against the event before the scan, so it can reference event fields. Empty means no range, which is what covers Shape 3.
- `refreshAfterMillis` — how long until a sliding-window ranking is queued for refresh. `-1` disables refresh, and refresh is skipped entirely for non-sliding declarations.
- `additionalProperties` — properties carried onto the rank row alongside the metric.

This replaces the earlier `topk: { topk, window: { from, to } }` shape. There is no `window` object; `ranges` plus `refreshAfterMillis` cover it.

### System Tables

Rank rows live in the `topk` database. The keys and property names are fixed in `AggregationConstants.Topk`.

Rank table

- Row key: `rankSource(database, table, topk, entity, dimensionValues)`, joined as `database|table|topk|entity|dim1|dim2|...`
- Properties: `metric` for the aggregated value, `additionalProperties` for the carried properties as one JSON string
- Index: `metric_desc`, so rows are already in top-K read order

Refresh queue table (`refresh`)

- Message key: `refreshKey(database, table, topk, entity, topkDimensionValue, dimensionValues)`, joined as `database|table|topk|entity|topkDimensionValue|dim1|...`
[...]

PR #494, opened 2026-08-19
Run a top-K aggregation and write the ranked rows (Step 3 of #386)

zipdoki:
## Summary

A group can declare a top-K, but nothing computes one. The declaration sits in metadata and no ranked row is ever written. This runs the aggregation.

`TopkAggregationHandler` scans the source group with the declaration's `ranges` interpolated against the event, sums `metric` per `(entity, dimension key)`, and writes rank rows keyed by `rankSource(...)` under the `metric_desc` index. When a declaration is a sliding window and `refreshAfterMillis > 0`, it also schedules a refresh
for the instant the bucket leaves the window. Consuming that refresh message is the next PR in the stack.

`POST /aggregations/v1/aggregate` drives it, taking a batch of edges:

```json
{
  "items": [
    {
      "database": "graph",
      "table": "likes",
      "edge": {
        "source": "...",
        "target": "...",
        "...": "..."
      }
    }
  ]
}
```

```json
{
  "items": [
    {
      "database": "graph",
      "table": "likes",
      "source": "...",
      "target": "...",
      "status": "OK",
      "error": null
    }
  ]
}
```

Status and error are reported per item, so one bad edge does not sink the rest of the batch.

Two bucket changes ride along because the window math depends on them. `Bucket` now has a test pinning how a date bucket rounds both a value and a range bound, which the refresh schedule reads directly, and its format precision is read in one place instead of being re-derived at each call site. Separately, rank keys escape `|` so
two rankings can never collide on one row.

Stacked on #493  — that needs to land first.

Part of #386.

## Test plan

- `./gradlew :core:test --tests '*BucketTest*'` — date bucket rounding for values and for range bounds
- `./gradlew :core:test --tests '*AggregationConstantsTest*'` — rank key composition and `|` escaping
- `./gradlew :engine:test --tests '*RankingInputsTest*'` — `ranges` interpolation, entity and dimension resolution
- `./gradlew :engine:test --tests '*TopkAggregationHandlerTest*'` — aggregation over the source group, rank row writes, refresh scheduling
- `./gradlew :server:test --tests '*MetadataAggControllerE2ETest*'` — `POST /aggregations/v1/aggregate` over HTTP
- `./gradlew spotlessKotlinCheck build` — formatting and full build

## AI Assistance

### DIFF
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/AggregationItemRequest.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/AggregationItemRequest.kt
@@ -1,0 +1,11 @@
+package com.kakao.actionbase.core.edge.payload
+
+data class AggregationItemRequest(
+    val items: List<AggregationItemPayload>,
+)
+
+data class AggregationItemPayload(
+    val database: String,
+    val table: String,
+    val edge: EdgePayload,
+)
--- a/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/AggregationResult.kt
+++ b/core/src/main/kotlin/com/kakao/actionbase/core/edge/payload/AggregationResult.kt
@@ -1,0 +1,10 @@
+package com.kakao.actionbase.core.edge.payload
+
+data class AggregationResult(
+    val database: String,
+    val table: String,
+    val source: String,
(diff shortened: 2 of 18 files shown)


zipdoki:
Optimistic Merge

PR #496, opened 2026-08-19
Recompute a ranking from a refresh message (Step 4 of #386)

zipdoki:
## Summary

The previous PR schedules a refresh for the instant a bucket leaves the sliding window, but nothing consumes it. A windowed ranking goes stale the moment its oldest bucket ages out, and stays stale. This consumes the refresh.

`POST /aggregations/v1/sweep` takes refresh messages and recomputes the ranking row each one names:

```json
{
  "items": [
    {
      "type": "TOPK",
      "item": {
        "database": "graph", "table": "likes", "topk": "daily_likes",
        "source": "...", "target": "...", "direction": "OUT",
        "entity": "...", "topkDimensionValue": "...", "dimensionValues": "",
        "ranges": "", "properties": {}, "refreshAt": 1755500000000
      }
    }
  ]
}
```

```json
{ "items": [ { "database": "graph", "table": "likes", "topk": "daily_likes", "entity": "...", "status": "OK", "error": null } ] }
```

`SweepItem` resolves `item` polymorphically off `type` through Jackson's `EXTERNAL_PROPERTY`, so a second aggregation kind adds a `SweepItemPayload` subtype and a `JsonSubTypes` entry rather than a second endpoint. `AggregationHandler` gains `sweep`, and `AggregationService` dispatches by `AggregationType` instead of fanning out to
every handler the way `aggregate` does — a refresh message already knows which kind it belongs to.

`TopkSweepItem` carries exactly what `refreshKey` encodes, plus `refreshAt`, so the consumer can rebuild the ranking without reading the queue row back.

Sweep is a write path and is rejected under read-only mode, alongside `POST /aggregations/v1/aggregate`.

Stacked on #494 — that needs to land first.

Part of #386.

## Test plan

- `./gradlew :engine:test --tests '*TopkAggregationHandlerTest*'` — recompute from a refresh message, including the rank row rewrite
- `./gradlew :engine:test --tests '*AggregationServiceTest*'` — dispatch to the handler named by `AggregationType`
- `./gradlew :server:test --tests '*MetadataAggControllerE2ETest*'` — `POST /aggregations/v1/sweep` over HTTP, per-item status and error
- `./gradlew :server:test --tests '*ReadOnlyRequestFilterTest*'` — sweep is blocked in read-only mode
- `./gradlew spotlessKotlinCheck build` — formatting and full build

## AI Assistance

- [x] This PR was written largely with AI assistance.
  - Tool / model: Claude Code (Opus 5)

PR #497, opened 2026-08-20
Read a materialized top-K by naming its entity and dimension values (Step 5 of #386)

zipdoki:
## Summary

Rank rows are written and kept fresh, but nothing can read them. Getting a ranking out means knowing the rank table's name, its index and how its row key is composed, then scanning it yourself. This exposes the read.

```
POST /aggregations/v1/databases/{database}/tables/{table}/topks/{topk}
```

The path names the source table and the declaration on it, never the rank table. `RankScan.from(...)` reads the declaration off the schema and resolves the rest — which table the rows live in, the `metric` index to walk, and the start key from `(database, table, topk, entity, dimensionValues)`. I kept the rank table out of the request on
purpose: a service that renames it should not have to tell its callers.

```json
{
  "entity": "user1",
  "dimensionValues": { "category": "fruit" },
  "limit": 10
}
```

```json
{
  "topks": [
    { "value": "item9", "metric": 42, "properties": { "brand": "acme" } },
    { "value": "item3", "metric": 17, "properties": { "brand": "acme" } }
  ],
  "count": 2
}
```

When the declaration is global, `entity` in the request is ignored — a global ranking is one row for everyone, so there is nothing to narrow. An unknown `topk` is a 400 naming it.

I put the read on `POST`, not `GET`. A ranking is picked by naming values, and a query string cannot carry those onto an immutable request. The downside is that a read stops looking like one to anything in front of the server — no caching, and `ReadOnlyRequestFilter` has to know about it by hand. It now matches on a path segment (`/topks/`)
as well as on a suffix, since this path ends in a variable. If the exception list grows, the filter should read the annotation instead of a name list.

Dimension values are positional against the declaration, and a value the request omits reads back as an empty string rather than as an error. So a partially specified request quietly reads the ranking split on empty values instead of the one the caller meant. I'll leave that validation for a follow-up.

Stacked on #496 — that needs to land first.

Part of #386.

## Test plan

- `./gradlew :engine:test --tests '*RankScanTest*'` — rank table resolution, key composition, global entity, unknown `topk`
- `./gradlew :server:test --tests '*MetadataAggQueryControllerE2ETest*'` — the endpoint over HTTP, per-entity and global rankings, dimension splits, paging
- `./gradlew :server:test --tests '*ReadOnlyRequestFilterTest*'` — the read passes in read-only mode

## AI Assistance


zipdoki:
Optimistic Merge

